PR #2, opened 2023-03-16
Improvements to actions, annotations and java.util.HashMap

dog-m:
Key changes:
- Improved type mappings.
- Improved annotations.
- Improved list actions (grouping + functionality enhancements).
- Added separate generator-specific annotations.
- Added unified interface descriptions **[draft]**.
- Added a few missing pieces into spec for `java.util.HashMap`.
- Added some *documentation* in the form of comments for annotations and actions.
- Fixed syntax incompatibilities in spec for `java.util.HashMap`.

**Note:** the specification is still missing `spliterator`-s for `key`-s, `value`-s and `entry`-es. If we decide to actually use the suggested `map<key, value>` internal type, then there will be a lot of changes.

### DIFF
--- a/spec/java-common.lsl
+++ b/spec/java-common.lsl
@@ -50,7 +50,7 @@ define action CALL(
 define action OBJECT_EQUALS(
         a: Object,
         b: Object
-    ): string;
+    ): boolean;
 
 define action OBJECT_TO_STRING(
         thing: Object
@@ -66,29 +66,33 @@ define action OBJECT_HASH_CODE(
 
 
 
-annotation Static ();
+annotation static ();
 
-annotation Final ();
+annotation final ();
 
-annotation Transient ();
+annotation transient ();
 
-annotation Private ();
+annotation private ();
 
-annotation PackagePrivate ();
+annotation packageprivate ();
 
-annotation Public ();
+annotation public ();
 
-annotation Extends (
+annotation extends (
     fullClassName: string;
 );
 
-annotation Implements (
+annotation implements (
     fullInterfaceNames: array<string>;
 );
 
-annotation Throws (
+annotation throws (
     exceptionTypes: array<string> = [];
 );
+
+annotation synchronized ();
+
+annotation varargs ();
 
 
 
--- a/spec/java-common.lsl
+++ b/spec/java-common.lsl
@@ -1,6 +1,11 @@
-libsl '1.1.0';
+libsl "1.1.0";
 
-library 'std:common' language 'Java' version '11' url '-';
+library "std" language "Java" version "11" url "-";
+
+
+// === IMPORTS ===
+
+import "generator-annotations.lsl";
 
 
 // === TYPES ===
@@ -12,7 +17,7 @@ typealias long    = int64;
 typealias float   = float32;
 typealias double  = float64;
 
-@Alias('java.lang.Object')
+@TypeMapping("java.lang.Object")
 typealias Object = *void;
 
 
--- a/spec/java-common.lsl
+++ b/spec/java-common.lsl
@@ -47,6 +47,7 @@ define action CALL(
 
 // "ad-hoc" solutions
 
+// the same as something like: return a == b || (a != null && a.equals(b))
 define action OBJECT_EQUALS(
         a: Object,
         b: Object
--- a/spec/java-common.lsl
+++ b/spec/java-common.lsl
@@ -2,33 +2,13 @@ libsl "1.1.0";
 
 library "std" language "Java" version "11" url "-";
 
-
-// === IMPORTS ===
-
+// DO NOT REMOVE!
 import "generator-annotations.lsl";
 
 
-// === TYPES ===
-
-
-@TypeMapping(builtin=true) typealias boolean = bool;
-@TypeMapping(builtin=true) typealias int     = int32;
-@TypeMapping(builtin=true) typealias long    = int64;
-@TypeMapping(builtin=true) typealias float   = float32;
-@TypeMapping(builtin=true) typealias double  = float64;
-@TypeMapping(builtin=true) typealias Object  = *void;
-
-
-// === CONSTANTS ===
-
-
-//val null: Object = 0;
-
-//val false: boolean = 0;
-//val true: boolean = 1;
-
 
 // === ACTIONS ===
+
 
 
 // specification development -related actions
@@ -79,3 +59,58 @@ define action OBJECT_TO_STRING(
 define action OBJECT_HASH_CODE(
         thing: Object
     ): int;
+
+
+
+// === ANNOTATIONS ===
+
+
+
+annotation Static ();
+
+annotation Final ();
+
+annotation Transient ();
+
+annotation Private ();
+
+annotation PackagePrivate ();
+
+annotation Public ();
+
+annotation Extends (
+    fullClassName: string;
+);
+
+annotation Implements (
+    fullInterfaceNames: array<string>;
+);
+
+annotation Throws (
+    exceptionTypes: array<string> = [];
+);
+
+
+
+// === CONSTANTS ===
+
+
+
+//val null: Object = 0;
+
+//val false: boolean = 0;
+//val true: boolean = 1;
+
+
+
+// === TYPES ===
+
+
+
+@TypeMapping(builtin=true) typealias boolean = bool;
+@TypeMapping(builtin=true) typealias int     = int32;
+@TypeMapping(builtin=true) typealias long    = int64;
+@TypeMapping(builtin=true) typealias float   = float32;
+@TypeMapping(builtin=true) typealias double  = float64;
+@TypeMapping(builtin=true) typealias Object  = *void;
+
--- a/spec/java-common.lsl
+++ b/spec/java-common.lsl
@@ -11,43 +11,27 @@ import "generator-annotations.lsl";
 // === TYPES ===
 
 
-typealias boolean = bool;
-typealias int     = int32;
-typealias long    = int64;
-typealias float   = float32;
-typealias double  = float64;
-
-@TypeMapping("java.lang.Object")
-typealias Object = *void;
+@TypeMapping(builtin=true) typealias boolean = bool;
+@TypeMapping(builtin=true) typealias int     = int32;
+@TypeMapping(builtin=true) typealias long    = int64;
+@TypeMapping(builtin=true) typealias float   = float32;
+@TypeMapping(builtin=true) typealias double  = float64;
+@TypeMapping(builtin=true) typealias Object  = *void;
 
 
 // === CONSTANTS ===
 
 
-val null: Object = 0;
+//val null: Object = 0;
 
-val false: boolean = 0;
-val true: boolean = 1;
+//val false: boolean = 0;
+//val true: boolean = 1;
 
 
 // === ACTIONS ===
 
 
-@NoReturn
-define action THROW_NEW(
-        exceptionType: string,
-        params: any[]
-    ): void;
-
-@NoReturn
-define action THROW_VALUE(
-        value: any
-    ): void;
-
-@NoReturn
-define action ERROR(
-        message: string
-    ): void;
+// specification development -related actions
 
 @NoReturn
 define action TODO(): void;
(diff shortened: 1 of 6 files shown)


spec/java/util/HashMap.lsl
```diff
@@ -75,87 +71,189 @@ automaton HashMap: int
...
+        assigns self.keys;
+        assigns self.values;
+        ensures self.length' < self.length;
+
+        result = action LIST_GET(values, index);
+
+        action LIST_REMOVE(keys, index);
+        action LIST_REMOVE(values, index);
+
+        length -= 1;
```
alexeev509, line 162: As I remember we have decided not to use increment and decrement at this moment... Or am I wrong ?

spec/java/util/HashMap.lsl
```diff
@@ -1,45 +1,41 @@
...
 )
 automaton HashMap: int
 (
-    var keys: list<Object>;
-    var values: list<Object>;
-    var length: int;
-    @Transient var modCounter: int;
+    var keys: list<Object> = null;
+    var values: list<Object> = null;
+    var length: int = 0;
```
alexeev509: Why is it not "@Transient" like in the original class ?

spec/java/util/HashMap.lsl
```diff
@@ -75,87 +71,189 @@ automaton HashMap: int
...
 
-    constructor HashMap (initialCapacity: int): void
-    requires initialCapacity >= 0;
-    assigns self.keys;
-    assigns self.values;
-    assigns self.length;
-    assigns self.modCounter;
-    ensures self.length = 0;
-    ensures self.modCounter = 0;
+    constructor HashMap (initialCapacity: int)
```
alexeev509, line 211: I understand that you use such require: "requires initialCapacity >= 0;". But in documentation if initialCapacity is less than 0, it throws "IllegalArgumentException": "Throws: IllegalArgumentException – if the initial capacity is negative.". 
If we will generate this code, I think we must throw "IllegalArgumentException". What do you think about this ? Maybe we must write something like this:
if (initialCapacity < 0)
{
     action THROW_NEW("java.lang.IllegalStateException", []);
}
or not ? 

how will we understand during generation that such an exception can fly out? is it important ?

spec/java/util/HashMap.lsl
```diff
@@ -75,87 +71,189 @@ automaton HashMap: int
...
+
+    proc _removeMapping (key: Object): Object
+    {
+        // side effects should be inferred from calls to other subroutines
+        ensures self.length' <= self.length;
+
+        val idx = action LIST_FIND(keys, key, 0, length, +1);
+        if (idx >= 0)
+        {
+            result = self._removeMapping(idx);
```
alexeev509, line 175: It willn't be problem for us to work with overloading ? As I remember we don't talk about this on our meetings... This isn't mistake, I really don't know...

spec/java-common.lsl
```diff
@@ -1,56 +1,116 @@
...
 define action CALL(
         callable: Object,
-        params: any[]
+        params: array<any>
     ): Object;
+
+
+// "ad-hoc" solutions
+
+define action OBJECT_EQUALS(
```
alexeev509, line 51: This action must return string ? Or boolean ?

spec/java/util/HashMap.lsl
```diff
@@ -314,122 +421,85 @@ automaton HashMap: int
...
-        val idx = action LIST_FIND(keys, key);
+        val idx = action LIST_FIND(keys, key, 0, length, +1);
         if (idx >= 0)
         {
             val oldValue = action LIST_GET(values, idx);
-            
-            // #problem
-            val isEqualValues = Objects.equals(value, oldValue);
+
+            val isEqualValues = Objects.equals(value, oldValue);  // #problem
```
alexeev509: Maybe here you must use new action:  "action OBJECT_EQUALS" ? Or not ?

spec/java/util/HashMap.lsl
```diff
@@ -314,122 +421,85 @@ automaton HashMap: int
...
         {
             val value = action LIST_GET(values, idx);
             // #problem
-            if (value == oldValue || (value != null && value.equals(oldValue)))
+            if (value == oldValue || (value != null && Objects.equals(value, oldValue)))
             {
                 action LIST_SET(values, idx, newValue);
 
-                self.updateModifications();
+                self._updateModifications();
```
alexeev509: I don't see change of the "modCount" field in this method... Correct me please if I'm wrong

spec/java/util/HashMap.lsl
```diff
@@ -314,122 +421,85 @@ automaton HashMap: int
...
-    // side effects should be inferred from calls to other methods
     {
-        val oldValue = self.get(key);
+        // side effects should be inferred from calls to other subroutines
+
+        val oldValue = self._getMappingOrDefault(key, null);
         if (oldValue != null)
         {
-            self.put(key, newValue);
+            self._setMapping(key, newValue);
```
alexeev509: I don't see change of the "modCount" field in this method... Correct me please if I'm wrong

spec/java/util/HashMap.lsl
```diff
@@ -510,87 +572,94 @@ automaton HashMap: int
 
     fun entrySet (): Set
     {
-        result = new HashMap_EntrySet(state=Initialized);
+        result = new HashMap_EntrySet(state=Initialized); // parent will be implicitly set to self
```
alexeev509, line 626: In the original class this method looks in such way:

 public Set<Map.Entry<K,V>> entrySet() {
        Set<Map.Entry<K,V>> es;
        return (es = entrySet) == null ? (entrySet = new EntrySet()) : es;
    }

As I understood it can be not only "new" object; How is this problem solved here ?

spec/java/util/HashMap.lsl
```diff
@@ -510,87 +572,94 @@ automaton HashMap: int
...
 
-    fun equals (other: Object): boolean
-    {
         // #problem
         action NOT_IMPLEMENTED();
     }
 
     @Private
-    @Throws(['IOException', 'ClassNotFoundException'])
+    @Throws(["IOException", "ClassNotFoundException"])
```
alexeev509: I think we must write full names of the classes: "java.io. IOException"; What do you think ?

spec/java/util/HashMap.lsl
```diff
@@ -510,87 +572,94 @@ automaton HashMap: int
...
+
+//        // #problem
+//        action FROM_BYTES(self, bytes);
     }
 }
 
 
 
 
+@From("HashMap")
```
alexeev509, line 709: Maybe use not "@From" but "@Inner" annotation ?

spec/java/util/HashMap.lsl
```diff
@@ -605,9 +674,16 @@ automaton HashMap_Values: int
...
+    fun contains(value: Object): boolean
     {
-        result = self.parent.containsValue(o);
+        if (self.parent.length == 0)
+        {
+            result = false;
+        }
+        else
+        {
+            result = action LIST_FIND(self.parent.values, value, 0, self.parent.length, +1);
```
alexeev509: LIST_FIND returns int; I think you must replace this code by this:

index =  action LIST_FIND(self.parent.values, value, 0, self.parent.length, +1);
result = index >= 0;

spec/java/util/HashMap.lsl
```diff
@@ -662,61 +742,75 @@ automaton HashMap_KeySet: int
...
+    fun contains (key: Object): boolean
     {
-        result = self.parent.containsKey(o);
+        if (self.parent.length == 0)
+        {
+            result = false;
+        }
+        else
+        {
+            result = action LIST_FIND(self.parent.keys, key, 0, self.parent.length, +1);
```
alexeev509: It doesn't return boolean

spec/java/util/HashMap.lsl
```diff
@@ -769,7 +862,323 @@ automaton HashMap_KeyIterator: int
...
+    {
+        result = self.parent.length;
+    }
+
+
+    fun clear (): void
+    {
+        assigns self.parent;
+
+        result = self.parent._clearMappings();
```
alexeev509: "result = " this is mistake, because type of the return value is "void"

spec/java/util/HashMap.lsl
dog-m: But not this form of increment/decrement. See item **17** in *Required Improvements*.
If, however, this item was rejected **without** my knowledge, then yes, indeed, there should used the complete form.

spec/java/util/HashMap.lsl
dog-m: If the serialization procedure will be described in specification, then, I suspect, we would have difficulties re-creating state: the lengths of lists would not be accessible. So it would be useful to leave it searializable, I guess.

spec/java/util/HashMap.lsl
dog-m: As I can infer from the sources of the parser library, each automaton has a list of functions (not a set). Thus I don't expect such code to fail parsing procedure or being inaccessible, during further processing (code generation, etc.).
I think we can left it as is and change it in the future when we would have a somewhat feature-complete working parser.

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java-common.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: Added `@CacheOnce` annotation.

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: A part of additional meaning will be lost here: the name `From` should indicate not only inclusion in another class but also refer to its parent from which we would be able to infer the type for the `parent` field.
I agree, may be there should be another name used here but I cannot think of other options.

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed

spec/java/util/HashMap.lsl
dog-m: Yes, you are correct. Fixed. No Idea how I missed that in other places also.

spec/java/util/HashMap.lsl
dog-m: Fixed for now

spec/java/util/HashMap.lsl
dog-m: If we completely ignore preconditions and postconditions in the generated executable code then yes, it should be explicitly written as throwing an exception.

dog-m: Fixed

spec/java/util/HashMap.lsl
```diff
@@ -314,283 +464,260 @@ automaton HashMap: int
...
         result = false;
 
-        val idx = action LIST_FIND(keys, key);
+        val idx = action LIST_FIND(keys, key, 0, length, +1);
         if (idx >= 0)
         {
             val value = action LIST_GET(values, idx);
             // #problem
-            if (value == oldValue || (value != null && value.equals(oldValue)))
+            if (value == oldValue || (value != null && Objects.equals(value, oldValue)))
```
alexeev509: This line must be replaced by action;

dog-m: Fixed

PR #3, opened 2023-03-28
java.util.ArrayList LibSL Spec

alexeev509:
SubList (and it's iterator) isn't ready. It was commented because we don't know what to do with it at this moment.

### DIFF
--- a/spec/java/util/ArrayList.lsl
+++ b/spec/java/util/ArrayList.lsl
@@ -133,123 +133,47 @@ automaton ArrayList: int (
         }
     }
 
-    sub grow (minCapacity: int): list<Object>
-    {
-
-    }
-
-
-    sub grow (): list<Object>
-    {
-
-    }
-
-
-    sub indexOfRange (o: Object, start: int, end: int): int
-    {
-
-    }
-
-
-    sub lastIndexOfRange (o: Object, start: int, end: int): int
-    {
-
-    }
-
-
-    sub elementData (index: int): Object
-    {
-
-    }
-
-
-    sub elementAt (es: list<Object>, index: int): Object
-    {
-
-    }
-
-
-    sub add (e: Object, elementData: list<Object>, s: int): void
-    {
-
-    }
-
-
-    sub equalsRange (other: List, from: int, to: int): boolean
-    {
-
-    }
-
-
-    sub equalsArrayList (other: ArrayList): boolean
-    {
-
-    }
-
-
-    sub checkForComodification (@Final expectedModCount: int): void
-    {
-
-    }
-
-
-    sub hashCodeRange (from: int, to: int):int
-    {
-
-    }
-
-
-    sub fastRemove (es: list<Object>, I: int): void
-    {
-
-    }
-
-
-    sub removeRange (fromIndex: int, toIndex): void
-    {
-
-    }
-
-
-    sub shiftTailOverGap (es: list<Object>, lo: int, hi: int): void
-    {
-
-    }
-
 
     sub rangeCheckForAdd (index: int): void
     {
-
+        if (index < 0 || index > length)
+        {
+            var message = "Index: " + action TO_STRING(index) + ", Size: " + action TO_STRING(length);
+            action THROW_NEW("java.lang.IndexOutOfBoundsException", [message]);
+        }
     }
 
 
-    sub outOfBoundsMsg (index: int): String
+    sub addAllElements (index:int, c:Collection): boolean
     {
+        modCount = modCount + 1;
 
+        //problem:
+        //we don't know how to avoid cycle i this method
+        action NOT_IMPLEMENTED();
+
+        length = length + c.length;
     }
 
-
-    sub outOfBoundsMsg (fromIndex: int, toIndex: int): String
+    sub subListRangeCheck (int fromIndex, int toIndex, int size): void
     {
+        if (fromIndex < 0)
+        {
+            var message: String = "fromIndex = " + action TO_STRING(fromIndex);
+            action THROW_NEW("java.lang.IndexOutOfBoundsException", [message]);
+        }
 
-    }
+        if (toIndex > size)
+        {
+            var message: String = "toIndex = " + action TO_STRING(toIndex);
+            action THROW_NEW("java.lang.IndexOutOfBoundsException", [message]);
+        }
 
-
-    sub batchRemove (c: Collection, complement: boolean, @Final from: int, @Final end: int)
-    {
-
-    }
-
-
-    sub removeIf (filter: Predicate, i: int, @Final end: int): boolean
-    {
-
-    }
-
-
-    sub replaceAllRange (operator: UnaryOperator, i: int, end: int): void
-    {
-
+        if (fromIndex > toIndex)
+        {
+            var message: String = "fromIndex(" + action TO_STRING(fromIndex) + ") > toIndex(" + action TO_STRING(toIndex) + ")";
+            action THROW_NEW("java.lang.IllegalArgumentException", [message]);
+        }
     }
 
 
@@ -349,129 +273,200 @@ automaton ArrayList: int (
 
     fun add (index: int, element: Object): void
     {
-
+         rangeCheckForAdd(index);
+         modCount = modCount + 1;
+         action LIST_INSERT_AT(storage, index, e);
+         length = length + 1;
     }
 
 
     fun remove (index: int): Object
     {
-
+        checkValidIndex(index);
+        result = action LIST_GET(storage, index);
+        action LIST_REMOVE(storage, index, 1);
+        modCount = modCount + 1;
+        length = length - 1;
     }
 
 
     fun equals (o: Object): boolean
     {
-
+        //Problem
+        //We don't know at this moment how create this method.
+        action NOT_IMPLEMENTED();
     }
 
 
     fun hashCode (): int
     {
-
+        // result = action OBJECT_HASH_CODE(self);
+        // #problem
+        action NOT_IMPLEMENTED();
     }
 
 
     fun remove (o: Object): boolean
     {
-
+        var index = action LIST_FIND(storage, o);
+        if (index == -1)
+        {
+            result = false;
+        }
+        else
+        {
+            action LIST_REMOVE(storage, index, 1);
+            result = true;
+        }
     }
 
 
     fun clear (): void
     {
-
+        action LIST_RESIZE(storage, 0);
+        length = 0;
+        modCount = modCount + 1;
     }
 
 
     fun addAll (c: Collection): boolean
     {
-
+        result = addAllElements(length, c);
     }
 
 
 
     fun addAll (index: int, c: Collection): boolean
     {
-
+        result = addAllElements(length, c);
     }
 
 
     fun removeAll (c: Collection): boolean
     {
+        //Problem.
+        //I want to create such action:
 
+        //define action SUBTRACTING_SETS(
+        //        aListSubtracting: array<any>,
+        //        aListDeductible: array<any>
+        //    ): list<any>;
+
+        //But Collection isn't List.
+        //But it has method: toArray();
+        //Maybe we must have two arrays as params ?
+
+        //storage = action SUBTRACTING_SETS(storage, c);
+
+        action NOT_IMPLEMENTED();
     }
 
 
     fun retainAll (c: Collection): boolean
     {
+        //Problem.
+        //I want to create such action:
 
+        //define action INTERSECTION(
+        //        aListSubtracting: array<any>,
+        //        aListDeductible: array<any>
+        //    ): list<any>;
+
+        //But Collection isn't List.
+        //But it has method: toArray();
+        //Maybe we must have two arrays as params ?
+
+        //storage = action INTERSECTION(storage, c);
+
+        action NOT_IMPLEMENTED();
     }
 
 
     @Private
     fun writeObject (s: ObjectOutputStream): void
     {
-
+        action NOT_IMPLEMENTED();
     }
 
 
     @Private
     fun readObject (s: ObjectInputStream): void
     {
-
+        action NOT_IMPLEMENTED();
     }
 
 
     fun listIterator (index: int): ListIterator
     {
+        rangeCheckForAdd(index);
 
+        result = new ListItr(state=Created,
+            cursor=index,
+            expectedModCount=modCount);
     }
 
 
     fun listIterator (): ListIterator
     {
-
+        result = new ListItr(state=Created,
+            cursor=index,
+            expectedModCount=modCount);
     }
 
 
     fun iterator (): Iterator
     {
-
+        result = new Itr(state=Created,
+            expectedModCount=modCount);
     }
 
 
     fun subList (fromIndex: int, toIndex: int): List
     {
-
+        subListRangeCheck(fromIndex, toIndex, length);
+        result = new Itr(state=Created,
+            //Think about THIS !
+            //TODO
+            start=fromIndex,
+            end=toIndex);
     }
 
 
     fun forEach (action: Consumer): void
     {
-
+        // #problem
+        action NOT_IMPLEMENTED();
     }
 
 
     fun spliterator (): Spliterator
     {
-
+        result = new ArrayListSpliterator(state=Initialized,
+            //This is right ? "parent=self"
+            parent=self,
+            origin = 0,
+            est=-1,
+            expectedModCount=0);
     }
 
 
     fun removeIf (filter: Predicate): boolean
     {
-
+        // #problem
+        action NOT_IMPLEMENTED();
     }
 
 
     fun replaceAll (operator: UnaryOperator): void
     {
-
+        // #problem
+        action NOT_IMPLEMENTED();
     }
 
 
     fun sort (c: Comparator): void
     {
-
+        // #problem
+        action NOT_IMPLEMENTED();
     }
 }
--- a/spec/java/util/ArrayList.lsl
+++ b/spec/java/util/ArrayList.lsl
@@ -177,6 +177,24 @@ automaton ArrayList: int (
     }
 
 
+    sub checkForComodification (expectedModCount: int): void
+    {
+        if (modCount != expectedModCount)
+        {
+            action THROW_NEW("java.util.ConcurrentModificationException", []);
+        }
+    }
+
+    sub deleteElement (index: int): Object
+    {
+        checkValidIndex(index);
+        result = action LIST_GET(storage, index);
+        action LIST_REMOVE(storage, index, 1);
+        modCount = modCount + 1;
+        length = length - 1;
+    }
+
+
     //methods
 
     fun trimToSize (): void
@@ -282,11 +300,7 @@ automaton ArrayList: int (
 
     fun remove (index: int): Object
     {
-        checkValidIndex(index);
-        result = action LIST_GET(storage, index);
-        action LIST_REMOVE(storage, index, 1);
-        modCount = modCount + 1;
-        length = length - 1;
+        result = deleteElement(index);
     }
 
 
@@ -424,7 +438,7 @@ automaton ArrayList: int (
     fun subList (fromIndex: int, toIndex: int): List
     {
         subListRangeCheck(fromIndex, toIndex, length);
-        result = new Itr(state=Created,
+        result = new SubList(state=Created,
             //Think about THIS !
             //TODO
             start=fromIndex,
@@ -470,3 +484,100 @@ automaton ArrayList: int (
         action NOT_IMPLEMENTED();
     }
 }
+
+
+
+@PackagePrivate
+@Implements("java.util.Iterator")
+@WrapperMeta(
+    src="java.util.ArrayList$Itr",
+    //Maybe will be another name of the dest class
+    dest="org.utbot.engine.overrides.collections.ArrayList$Itr",
+    matchInterface=true)
+automaton Itr: int (
+    cursor: int,
+    lastRet: int = -1,
+    expectedModCount: int
+) {
+
+     initstate Initialized;
+
+     shift Allocated -> Initialized by [
+             Itr()
+     ];
+
+     shift Initialized -> self by [
+        // read operations
+        hasNext,
+
+        // write operations
+        next,
+        remove,
+        forEachRemaining
+     ]
+
+
+     // constructors
+
+
+     //Maybe here is needed @DefaultModifier ?
+     constructor Itr()
+     {
+     }
+
+
+     // methods
+
+
+     fun hasNext (): boolean
+     {
+         result = cursor != self.parent.length;
+     }
+
+
+     fun next (): Object
+     {
+         self.parent.checkForComodification(expectedModCount);
+         var i = cursor;
+
+         if (i >= parent.length)
+         {
+             action THROW_NEW("java.util.NoSuchElementException", []);
+         }
+
+         //Problem
+         //I don't know what to do with ConcurrentModificationException(); ?
+
+         cursor = i + 1;
+         lastRet = i;
+         result = action LIST_GET(self.parent.storage, lastRet);
+     }
+
+
+     fun remove (): void
+     {
+         if (lastRet < 0)
+         {
+             action THROW_NEW("java.lang.IllegalStateException", []);
+         }
+
+         self.parent.checkForComodification(expectedModCount);
+
+         //Problem
+         //What i must to do with try-catch in this method ?
+
+         self.parent.deleteElement(lastRet);
+         cursor = lastRet;
+         lastRet = -1;
+         expectedModCount = self.parent.modCount;
+     }
+
+
+     fun forEachRemaining (action: Consumer): void
+     {
+         action NOT_IMPLEMENTED();
+     }
+
+}
+
+
--- a/spec/java/util/ArrayList.lsl
+++ b/spec/java/util/ArrayList.lsl
@@ -581,3 +581,87 @@ automaton Itr: int (
 }
 
 
+
+@PackagePrivate
+@Extends("java.util.ArrayList$Itr")
+@Implements("java.util.Iterator")
+@WrapperMeta(
+    src="java.util.ArrayList$Itr",
+    //Maybe will be another name of the dest class
+    dest="org.utbot.engine.overrides.collections.ArrayList$Itr",
+    matchInterface=true)
+automaton ListItr: int (
+    cursor: int,
+    lastRet: int = -1,
+    expectedModCount: int
+) {
+
+    initstate Initialized;
+
+    shift Allocated -> Initialized by [
+        ListItr(int)
+    ];
+
+    shift Initialized -> self by [
+       // read operations
+       hasPrevious,
+       nextIndex,
+
+       // write operations
+       previous,
+       set,
+       add
+    ]
+
+    //constructors
+
+
+    constructor ListItr(index: int)
+    {
+        //Problem
+        //How invoke super() ??
+
+    }
+
+
+    //methods
+
+
+    fun hasPrevious (): boolean
+    {
+
+    }
+
+
+    fun nextIndex (): int
+    {
+
+    }
+
+
+    fun previousIndex (): int
+    {
+
+    }
+
+
+    fun previous (): Object
+    {
+
+    }
+
+
+    fun set (e: Object): void
+    {
+
+    }
+
+
+    fun add (e: Object): void
+    {
+
+    }
+
+}
+
+
--- a/spec/java/util/ArrayList.lsl
+++ b/spec/java/util/ArrayList.lsl
@@ -124,7 +124,7 @@ automaton ArrayList: int (
 
     //subs
 
-    sub checkValidIndex (index: int, length: int): void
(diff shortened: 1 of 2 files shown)


spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+include java.util.stream.Stream;
+
+// automata
+
+@extends("java.util.AbstractList")
+@implements(["java.util.List", "java.util.RandomAccess","java.lang.Cloneable","java.io.Serializable"])
+@WrapperMeta(
+    src="java.util.ArrayList",
+    dst="ru.spbpu.libsl.overrides.collections.ArrayList",
+    forceMatchInterfaces=true)
```
dog-m: What is the point of forcing interfaces from the source class if you explicitly specified them using the `@implements` annotation?

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+automaton ArrayList: int (
+    @private @static @final var serialVersionUID: long = 8683452581122892189,
+    @transient var storage: List<Object> = null,
```
dog-m: The `storage` variable should not be `transient`. If I understand correctly the contents of this "field" should be saved and restored during serialization process. So it should be stored and, thus, not have the attribute/annotation `@transient`.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        trimToSize
+    ];
+
+    //constructors
+
+    constructor ArrayList (initialCapacity: int)
+    {
+        if (initialCapacity >= 0)
+        {
+            action LIST_RESIZE(storage, initialCapacity);
```
dog-m: As we are **not using real arrays** here, there should be no use for things like *capacity*. Only *size* and *items* actually matter.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        }
+        else
+        {
+            var message = "Illegal Capacity: " + action OBJECT_TO_STRING(initialCapacity);
```
dog-m: I'm not sure it would be allowed to use `action`-s with strings directly. I suggest using separate variable for the result of an `action`.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+
+    constructor ArrayList ()
+    {
+        action LIST_RESIZE(storage, 0);
+    }
+
+
+    constructor ArrayList (c: Collection)
+    {
+        //I suppose that "c" is another automaton and it has "toArray" sub (or with another sub name).
```
dog-m: We cannot assume that the supplied collection is an automaton of any known type. The `Collection` is just an interface. Thus, it is possible that the user can create its own very-specific-collection that satisfies the `Collection` interface contract and pass it into this constructor.
So, **at most** we should rely on and call only existing interface methods.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        //That's why we can invoke this.
+        //action ARRAY_TO_LIST(c.toArray(), storage);
+        // #problem:
+        //In the next code of the original class can be such situation:  https://bugs.openjdk.java.net/browse/JDK-6260652
```
dog-m: I don't think it is relevant here as this collection (at least the original one) operates using only `Object` references.
Also, it was fixed in version 9 as stated on the provided bug-tracker page.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    {
+        modCount = modCount + 1;
+
+        // #problem:
+        //we don't know how to avoid cycle in this method;
+        //for e in c:
+        //   storage.add(e);
+        action NOT_IMPLEMENTED();
+
+        length = length + c.size();
```
dog-m: It is not possible at the moment to make calls to interface methods.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    }
+
+    proc _subListRangeCheck (int fromIndex, int toIndex, int size): void
```
dog-m: Incorrect syntax for parameters here.

Also, it would be very helpful if there would be a doc-string or comment describing the meaning of such operation. Something like this:
```
// checks range [from, to) against [0, size)
```

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        action LIST_SET(storage, index, element);
+    }
+
+
+    //methods
+
+    fun trimToSize (): void
+    {
+        modCount = modCount + 1;
+        //As we work with List we mustn't resize it... Or not ?
```
dog-m: Absolutely correct.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    }
+
+
+    fun ensureCapacity (minCapacity: int): void
+    {
+        modCount = modCount + 1;
+        //As we work with List we mustn't resize it... Or not ?
```
dog-m: Absolutely correct.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    fun indexOf (o: Object): int
+    {
+        result = action LIST_FIND(storage, o, 0, length, +1);
+    }
+
+
+
+    fun lastIndexOf (o: Object): int
+    {
+        result = action LIST_FIND(storage, o, 0, length, -1);
```
dog-m: Incorrect starting position for search. It should be something like `length-1`.
Also the end position should be something like `-1`.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    }
+
+
+    fun clone (): Object
+    {
+        storageCopy = action LIST_DUP(storage);
```
dog-m: Missing local variable declaration.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    }
+
+
+    fun remove (index: int): Object
+    {
+        result = _deleteElement(index);
+    }
+
+
+    fun equals (other: Object): boolean
```
dog-m, line 326: You can see the example [here](https://github.com/vpa-research/jsl-spec/blob/patches-mikhail/spec/java/util/Optional.lsl#L374).

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+            // #problem
+            //We don't know at this moment how create this method.
+            action NOT_IMPLEMENTED();
+        }
+    }
+
+
+    fun hashCode (): int
+    {
+        // result = action OBJECT_HASH_CODE(self);
```
dog-m: I don't see a problem here as we have the `OBJECT_HASH_CODE` action.
I think it would be just `hash(storage)`, doesn't it?

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        //define action SUBTRACTING_SETS(
+        //        aListSubtracting: array<any>,
+        //        aListDeductible: array<any>
+        //    ): list<any>;
+
+        //But Collection isn't List.
+        //But it has method: toArray();
+        //Maybe we must have two arrays as params ?
+
+        //storage = action SUBTRACTING_SETS(storage, c);
```
dog-m: A foreign `Collection` is not a `list` nor an `array`. It's an *interface*.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        //define action INTERSECTION(
+        //        aListSubtracting: array<any>,
+        //        aListDeductible: array<any>
+        //    ): list<any>;
+
+        //But Collection isn't List.
+        //But it has method: toArray();
+        //Maybe we must have two arrays as params ?
+
+        //storage = action INTERSECTION(storage, c);
```
dog-m: A foreign `Collection` is not a `list` nor an `array`. It's an *interface*.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    {
+        result = new ListItr(state=Created,
+            cursor=0,
+            expectedModCount=modCount);
+    }
+
+
+    fun iterator (): Iterator
+    {
+        result = new Itr(state=Created,
```
dog-m: I don't see why we should use two different classes/automata where one is just an extension of another.
Why not just merge them in one and use only `ListItr`?

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        // #problem
+        action NOT_IMPLEMENTED();
+    }
+
+
+    fun spliterator (): Spliterator
+    {
+        result = new ArrayListSpliterator(state=Initialized,
+            //This is right ? "parent=self"
+            parent=self,
```
dog-m: This is correct but not required. It should be set automatically.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+            origin = 0,
+            est=-1,
+            expectedModCount=0);
+    }
+
+
+    fun removeIf (filter: Predicate): boolean
+    {
+        // #problem
+        //We don't know how to work with lambda.
```
dog-m: Incorrect. We do have an action `CALL` that is designed to make calls to functional Java interfaces (i.e. where only a single abstract method resides).

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+
+        modCount = modCount + 1;
+    }
+}
+
+
+
+@packagePrivate
+@implements("java.util.Iterator")
+@WrapperMeta(
```
dog-m: This should not be a wrapper as it is just an internal/private class.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    //Maybe will be another name of the dst class
+    dst="ru.spbpu.libsl.overrides.collections.ArrayList_Itr",
+    forceMatchInterfaces=true)
+automaton Itr: int (
+    var cursor: int,
+    var lastRet: int = -1,
+    var expectedModCount: int
+) {
+
+    initstate Initialized;
```
dog-m: The `Allocated` state is missing but being involved in shifts.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        self.parent._checkForComodification(expectedModCount);
+        var i = cursor;
+
+        if (i >= parent.length)
+        {
+            action THROW_NEW("java.util.NoSuchElementException", []);
+        }
+
+        // #problem
+        //I don't know what to do with ConcurrentModificationException(); ?
```
dog-m: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    {
+        if (lastRet < 0)
+        {
+            action THROW_NEW("java.lang.IllegalStateException", []);
+        }
+
+        self.parent._checkForComodification(expectedModCount);
+
+        // #problem
+        //What i must to do with try-catch in this method ?
```
dog-m: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    }
+
+}
+
+
+
+@packagePrivate
+@extends("java.util.ArrayList$Itr")
+@implements("java.util.Iterator")
+@WrapperMeta(
```
dog-m: This should not be a wrapper as it is just an internal/private class.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    {
+        action NOT_IMPLEMENTED();
+    }
+
+}
+
+
+
+@packagePrivate
+@extends("java.util.ArrayList$Itr")
```
dog-m: This is not possible as `java.util.ArrayList.Itr` is an internal/private class thus inaccessible from other packages.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+@implements("java.util.Iterator")
```
dog-m: As I can see in the original class, there should be `java.util.ListIterator` instead. Also missing square brackets.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    //Maybe will be another name of the dst class
+    dst="ru.spbpu.libsl.overrides.collections.ArrayList_ListItr",
+    forceMatchInterfaces=true)
+automaton ListItr: int (
+    var cursor: int,
+    var lastRet: int = -1,
+    var expectedModCount: int
+) {
+
+    initstate Initialized;
```
dog-m, line 584: The `Allocated` state is missing but being involved in shifts.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+       add
+    ]
+
+    //constructors
+
+
+    constructor ListItr(index: int)
+    {
+        // #problem
+        //How invoke super() ??
```
dog-m: Automata does not have inheritance yet so no calls to "super-automaton".

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        self.parent._checkForComodification(expectedModCount);
+        var i = cursor;
+
+        if (i >= parent.length)
+        {
+            action THROW_NEW("java.util.NoSuchElementException", []);
+        }
+
+        // #problem
+        //I don't know what to do with ConcurrentModificationException(); ?
```
dog-m, line 654: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        self.parent._checkForComodification(expectedModCount);
+        var i = cursor - 1;
+
+        if (i < 0)
+        {
+            action THROW_NEW("java.util.NoSuchElementException", []);
+        }
+
+        // #problem
+        //I don't know what to do with ConcurrentModificationException(); ?
```
dog-m, line 674: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+    {
+        if (lastRet < 0)
+        {
+            action THROW_NEW("java.lang.IllegalStateException", []);
+        }
+
+        self.parent._checkForComodification(expectedModCount);
+
+        // #problem
+        //What i must to do with try-catch in this method ?
```
dog-m, line 693: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
```
dog-m, line 713: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        self.parent._setElement(lastRet, e);
+    }
+
+
+    fun add (e: Object): void
+    {
+        self.parent._checkForComodification(expectedModCount);
+
+        // #problem
+        //What i must to do with try-catch in this method ?
```
dog-m, line 725: This is an example of a non-deterministic behavior. Place `NOT_IMPLEMENTED` action call here.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        action NOT_IMPLEMENTED();
+    }
+
+}
+
+
+
+@final
+@implements("java.util.Spliterator")
+@WrapperMeta(
```
dog-m: This should not be a wrapper as it is just an internal/private class.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+       // write operations
+       trySplit,
+       tryAdvance,
+       forEachRemaining
+    ]
+
+    //constructors
+
+
+    constructor ArrayListSpliterator (origin: int, fence: int, expectedModCount: int)
```
dog-m, line 775: It is not possible (except reflection) to call the constructors of this class outside of the `ArrayList`.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+       forEachRemaining,
+       previous,
+       set,
+       add
+    ]
+
+    //constructors
+
+
+    constructor ListItr(index: int)
```
dog-m, line 610: It is not possible (except reflection) to call the constructors of this class outside of the `ArrayList`.

spec/java/util/ArrayList.lsl
```diff
@@ -0,0 +1,1205 @@
...
+        // write operations
+        next,
+        remove,
+        forEachRemaining
+    ]
+
+
+    // constructors
+
+    constructor Itr()
```
dog-m: It is not possible (except reflection) to call the constructors of this class outside of the `ArrayList`.

spec/list-actions.lsl
```diff
@@ -10,6 +10,13 @@ define action LIST_COPY(
         to: int
     ): list<any>;
 
+define action LIST_TO_ARRAY(
```
dog-m, line 16: There should be a comment on why there are both *input* and *output* arrays.

dog-m (changes requested):
I strongly suggest merging `Itr` and `ListItr` into a single automaton and using it instead of separate `Itr`.

spec/java/util/ArrayList.lsl
alexeev509: @dog-m But why this is "transient" in the original class ? 
`transient Object[] elementData;`

[Why does ArrayList use transient storage?](https://stackoverflow.com/questions/9848129/why-does-arraylist-use-transient-storage)

As I understood we have fixed size of list (storage.size() == length == true), that's why we can make it NOT transient... Not like in the original ArrayList class. This is right ?

spec/java/util/ArrayList.lsl
alexeev509: @dog-m As I understood you it must looks in such way:

```
    constructor ArrayList (initialCapacity: int)
    {
        if (initialCapacity < 0)
        {
            var initCapacity = action OBJECT_TO_STRING(initialCapacity);
            var message = "Illegal Capacity: " + initCapacity;
            action THROW_NEW("java.lang.NoSuchElementException", [message]);
        }
    }
```

spec/java/util/ArrayList.lsl
alexeev509: @dog-m When I created review of the "HashMap" I saw in the one of there constructors such codeline:

```
constructor HashMap (other: Map)
    {
        ...
        val otherSize = other.size();
```

That's why I thought that this was suitable. It was mistake ?

spec/java/util/ArrayList.lsl
alexeev509: @dog-m As I understood you - I must add annotation "@private" ?

spec/list-actions.lsl
alexeev509: @dog-m In such way: 
`// We have input and output arrays, because we have method with such signature: public <T> T[] toArray(T[] a) `

?

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Was added;

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: @dog-m I agree with you; Was fixed;

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Parent was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Comment was deleted;

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: @dog-m Yes, but it has cycle. It must compare every element of the "self" and "other" arrays. That's why I wrote "action NOT_IMPLEMENTED();"; I'm right or not ?

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
dog-m: If we imagine the concept of `list` in relation to not-so-long-ago agreed piece of *target API*, then it does have the `size: () -> int` method. But for every spec we have created to this moment we used additional *variable* to store the size of the list. And, yes, at some point we did have an `action LIST_SIZE` or something similar. But calling every time such expensive method (involving unknown amount of some SMT interfacing stuff) is not a great idea.

In short, each list object have some internal knowledge about its size, so no need in storing additional `length`s or `size`es.

And, no, the approach from the original ArrayList does not apply to our specification because we use dynamic lists instead of fixed arrays. So no "excess empty tail" problem.

In general, I have no clear answer on how to do serialization yet because we haven't deeply researched the related machinery yet (this includes *special private methods*).

I insist on making this field/variable and `length` both to be non-transient just in case we decide to use some automatic serialization mechanism.

spec/java/util/ArrayList.lsl
dog-m: Yes, but there should also be another exception type: `java.lang.IllegalArgumentException`

spec/java/util/ArrayList.lsl
dog-m: Well noted! Yes, it was a mistake from my side. Or at least I forgot to put it into a comment block.
So, I suggest using `action NOT_IMPLEMENTED()` and a multi-line comment block `/* */` with rough implementation there and in similar places.
Also, feel free to apply it to the `HashMap` if you want.

spec/java/util/ArrayList.lsl
dog-m: What is stopping us from calling `OBJECT_EQUALS` on lists if we already sure the supplied object `isSameType` as this automaton?
We know for sure it has exactly the same automata variables. So why not just compare them variable-by-variable directly? (starting from `length`s)

spec/java/util/ArrayList.lsl
dog-m: My bad. If we are removing from the *whole list* of stored items, then yes, the loop is unavoidable until we would have LibSL-level lambdas and a loop-like action.

spec/java/util/ArrayList.lsl
dog-m: I don't think there should even be constructors in the first place inside of such internal classes/automata. We have no way of making calls to it from LibSL perspective (it is a public/`fun`-like function after all).

I don't know how common it is to call such constructors (in the original) using reflection in production code. I would consider bad/dangerous such behavior leaving the body of similar constructors of just a single call to `action ERROR`. Something like this:
```prolog
constructor ArrayListSpliterator (origin: int, fence: int, expectedModCount: int)
{
    action ERROR("Dangerous behavior, IDK");
}
```

spec/list-actions.lsl
dog-m: No, at least something like:
```scala
// Copies contents of the list into the provided array. Allows auto-resizing.
// Example behavior from Java: String[] x = arrList.toArray(new String[0]);
```

alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: It was commented; 

```
 //At this moment we can't work with Collection, because this is interface.
 //length = length + c.size();
```

spec/java/util/ArrayList.lsl
alexeev509: Fixed

spec/java/util/ArrayList.lsl
alexeev509: Fixed

dog-m (approved):

PR #12, opened 2023-09-13
Spec for `java.lang.StringBuilder`

alexeev509:
Was created StringBuilder.lsl spec;


spec/java/util/StringBuilder.lsl
```diff
@@ -0,0 +1,37 @@
+///#! pragma: non-synthesizable
```
dog-m: This comment should be exactly like this (just like in `ArrayList.lsl`):
```java
//#! pragma: non-synthesizable
```
This will exclude this file from the discovery list of the translator and prevent type re-declaration during parsing.

Also, **invalid file location**: expecting something like `spec/java/lang/StringBuilder.lsl` (original file is located in `lang` package).

spec/java/util/StringBuilder.lsl
```diff
@@ -0,0 +1,37 @@
...
+
+import java.common;
+import java/lang/_interfaces;
+import java/io/_interfaces;
+
+
+// primary types
+
+@implements("java.io.Serializable")
+@implements("java.lang.Comparable<StringBuilder>")
```
dog-m: Generic type parameter can be removed.

spec/java/util/StringBuilder.lsl
```diff
@@ -0,0 +1,37 @@
...
+    for Object
+{
+    //@private @static @final var serialVersionUID: long = 4383685877147921099;
+}
+
+
+@implements("java.io.Serializable")
+@implements("java.lang.Comparable")
+@implements("java.lang.CharSequence")
+@public @final type StringBuffer
```
dog-m: Why `StringBuffer` is here and not in its own separate file?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/lang/StringBuilder.java";
+
+// imports
+
+import java/util/StringBuilder;
+
+
+// === CONSTANTS ===
+
+val MAX_CODE_POINT: int = 1114111;
```
dog-m: If these constants are not being used by other automata, they can be moved into the origin type declaration.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    proc _appendCharSequence (seq: CharSequence): void
+    {
+        if (seq == null)
+            seq = "null";
+
+        val seqLength: int = action CALL_METHOD(seq, "length", []);
+        this.length += seqLength;
+
+        var i: int = 0;
+        action LOOP_FOR(i, 0, seqLength, +1, _appendCharSequence_loop(i, seq));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        this.storage += action OBJECT_TO_STRING(currentChar);
+    }
+
+
+    proc _appendString (str: String): void
+    {
+        if (str == null)
+            str = "null";
+        this.storage += str;
+        this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: Why not increment `this.length` by the result of `length()` on `str`? It is definitely shorter than it plus `this.storage`.
Also, when the supplied string is known to be `null`, then its `length` is already known (4 characters).

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+
+    proc _delete (start: int, end: int): void
+    {
+        val len: int = this.length - end + start;
+        var newString: array<char> = action ARRAY_NEW("char", len);
+
+        var i: int = 0;
+        var currentIndex: int = 0;
+        action LOOP_FOR(i, 0, this.length, +1, _deleteCharAt_loop(i, start, end, currentIndex, newString));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        this.storage = action OBJECT_TO_STRING(newString);
+        this.length = len;
+    }
+
+
+    @Phantom proc _deleteCharAt_loop (i: int, start: int, end:int, currentIndex:int, newString: array<char>): void
+    {
+        if (i < start || i >= end)
```
dog-m: I don't think this will be more efficient than using two loops copying characters *before* and *after* the specified segment: [0..start] + [end..length].

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        _checkRange(start, end, len);
+
+        val countInsertedElements: int = end - start;
+        val newStr: array<char> = action ARRAY_NEW("char", this.length + countInsertedElements);
+
+        var i: int = 0;
+        var currentIndex: int = start;
+        val endIndex: int = dstOffset + countInsertedElements;
+
+        action LOOP_FOR(i, 0, this.length, +1, _insertCharSequence_loop(i, dstOffset, endIndex, currentIndex, newStr, s));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        this.storage = action OBJECT_TO_STRING(newStr);
+        this.length += countInsertedElements;
+    }
+
+
+    @Phantom proc _insertCharSequence_loop(i: int, dstOffset: int, endIndex: int, currentIndex:int, newStr: array<char>, s: CharSequence): void
+    {
+        if (i < dstOffset)
```
dog-m: Will it be much simpler for the USVM to process three simple loops with *known* ranges or a single one with three branching statements on every iteration?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        {
+            val currentChar_1: char = action CALL_METHOD(this.storage, "charAt", [i]);
```
dog-m: Why not use direct assignment? Local variable `currentChar_1` isn't used anywhere else as I can see.
This relates to `currentChar_2` and `currentChar_3` also.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+
+    proc _substring(start: int, end: int): String
+    {
+        _checkRangeSIOOBE(start, end, this.length);
+        val sizeNewString: int = end - start;
+        val newStr: array<char> = action ARRAY_NEW("char", sizeNewString);
+
+        var i: int = 0;
+        action LOOP_FOR(i, start, end, +1, _newSubString_loop(i, newStr));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    @Phantom proc _newSubString_loop (i: int, newStr: array<char>): void
+    {
+        newStr[i] = action CALL_METHOD(this.storage, "charAt", [i]);
+    }
+
+    // constructors
+
+    constructor *.StringBuilder (@target self: StringBuilder)
+    {
+
```
dog-m: Empty constructor. Please add a comment with a clarification on why it is like that.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        if (str == null)
+            _throwNPE();
+
+        _appendString(str);
+    }
+
+
+    constructor *.StringBuilder (@target self: StringBuilder, capacity: int)
+    {
+
```
dog-m: Empty constructor. Please add a comment with a clarification on why it is like that.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, seq: CharSequence, start: int, end: int): StringBuilder
+    {
+        if (seq == null)
+            seq = "null";
+        val seqLength: int = action CALL_METHOD(seq, "length", []);
```
dog-m: Why the USVM should call a `length()` method on a constant string ("`null`")?
If we know the result already, why not just specify it there?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        _checkRange(start, end, seqLength);
+        this.length += end - start;
+        var i: int = 0;
+        action LOOP_FOR(i, start, end, +1, _appendCharSequence_loop(i, seq));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    {
+        if (obj == null)
+        {
+            this.storage += "null";
+            this.length += 4;
+        }
+        else
+        {
+            this.storage += action OBJECT_TO_STRING(obj);
+            this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: The string representation of the supplied object `obj` combined with size of the `this.storage` will be greater than just on itself.
I suggest using temporary variable for storing the `obj` string representation and **then** adding its `length()` value to `this.length`.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    {
+        if (sb == null)
+        {
+            this.storage += "null";
+            this.length += 4;
+        }
+        else
+        {
+            // That's right for StringBuffer ?
+            this.storage += action OBJECT_TO_STRING(sb);
```
dog-m: We should check to see if the provided `StringBuffer` **has** this automaton "attached" to it. This can help us access its internal state (`storage` and `length`) without expensive calls.

I should also suggest using shorter `String` source before calling `length()`. See the comment above.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+            this.storage += "false";
+            this.length += 5;
+        }
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, c: char): StringBuilder
+    {
+        // That's right for char ?
```
dog-m: Then what is the point in having `this.length`? What does it for?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        var i: int = 0;
+        action LOOP_FOR(i, 0, strSize, +1, _appendCharsArray_loop(i, str));
+        result = self;
+    }
+
+
+    @Phantom proc _appendCharsArray_loop (i: int, str: array<char>): void
+    {
+        var currentChar: char = str[i];
+        this.storage += action OBJECT_TO_STRING(currentChar);
```
dog-m: This operator is using `String::concat(String)` under the hood. So calling it with just a single character is very inefficient in terms of computational complexity.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, str: array<char>): StringBuilder
+    {
+        val strSize: int = action ARRAY_SIZE(str);
+        this.length += strSize;
+        var i: int = 0;
+        action LOOP_FOR(i, 0, strSize, +1, _appendCharsArray_loop(i, str));
```
dog-m: Why not just `OBJECT_TO_STRING` and `+=`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    fun *.append (@target self: StringBuilder, str: array<char>, offset: int, len: int): StringBuilder
+    {
+        val end: int = offset + len;
+        val strSize: int = action ARRAY_SIZE(str);
+        _checkRange(offset, end, strSize);
+        var i: int = 0;
+        action LOOP_FOR(i, 0, strSize, +1, _appendCharsArray_loop(i, str));
```
dog-m: Where does the `offset` have its role here?

Also suggesting copying a segment of the supplied array and turning it into a `String`. Then adding it into `this.storage` **without** asking its `length()`.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        this.length += strSize;
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, d: double): StringBuilder
+    {
+        this.storage += action OBJECT_TO_STRING(d);
+        this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: `length()` of the `double` in most cases shorter than `this.storage`. And requires less computational/analytical resources.
*(a hint on using local string variable)*

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, f: float): StringBuilder
+    {
+        this.storage += action OBJECT_TO_STRING(f);
+        this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: High computational complexity. Please, optimize a bit.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, i: int): StringBuilder
+    {
+        this.storage += action OBJECT_TO_STRING(i);
+        this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: High computational complexity. Please, optimize a bit.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.append (@target self: StringBuilder, lng: long): StringBuilder
+    {
+        this.storage += action OBJECT_TO_STRING(lng);
+        this.length = action CALL_METHOD(this.storage, "length", []);
```
dog-m: High computational complexity. Please, optimize a bit.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+            val curChar: char = codePoint as char;
+            this.storage += action OBJECT_TO_STRING(curChar);
+            this.length += 1;
+        }
+        else if (_isValidCodePoint(codePoint))
+        {
+            val firstChar: char = _lowSurrogate(codePoint);
+            val secondChar: char = _highSurrogate(codePoint);
+            this.storage += action OBJECT_TO_STRING(firstChar);
+            this.storage += action OBJECT_TO_STRING(secondChar);
```
dog-m: I think using an array of 2 elements will be a bit better than creating 2 strings from single characters and then merging both into `this.storage`.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
+///#! pragma: non-synthesizable
```
dog-m: **Invalid file location**: expecting something like `spec/java/lang/StringBuilder.main.lsl` (original file is located in `lang` package).

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        offsetByCodePoints,
+        hashCode,
+        codePoints,
+        chars,
+    ];
+
+    // internal variables
+
+    var storage: String = "";
+    var length: int = 0;
```
dog-m: Please add a comment on what does this internal variable counting? Characters? Bytes?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+    fun *.compareTo (@target self: StringBuilder, another: StringBuilder): int
+    {
+        if (another == self)
+        {
+            result = 0;
+        }
+        else
+        {
+            val anotherString: String = action OBJECT_TO_STRING(another);
```
dog-m: Why not access its state directly? *(when it IS known to have this automaton attached)*

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, s: CharSequence): StringBuilder
+    {
+        _checkOffset(dstOffset);
+
+        if (s == null)
+            s = "null";
+
+        val len: int = action CALL_METHOD(s, "length", []);
```
dog-m: String `length()` is known.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        _delete(index, index + 1);
+
+        result = self;
+    }
+
+
+    fun *.indexOf (@target self: StringBuilder, str: String): int
+    {
+        result = action DEBUG_DO("this.storage.indexOf(str)");
```
dog-m: A candidate for a new method inside of `String` or `CharSequence` type declaration.
And then calling it with `CALL_METHOD`.
Or `indexOf(str, 0)`.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    fun *.indexOf (@target self: StringBuilder, str: String, fromIndex: int): int
+    {
+        result = action DEBUG_DO("this.storage.indexOf(str, fromIndex)");
```
dog-m: A candidate for a new method inside of `String` or `CharSequence` type declaration.
And then calling it with `CALL_METHOD`.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, s: CharSequence, start: int, end: int): StringBuilder
+    {
+        _checkOffset(dstOffset);
+
+        if (s == null)
+            s = "null";
+
+        val len: int = action CALL_METHOD(s, "length", []);
```
dog-m: String `length()` is known.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, obj: Object): StringBuilder
+    {
+        _checkOffset(dstOffset);
+
+        var s: String = "null";
```
dog-m: String `length()` is known.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, s: String): StringBuilder
+    {
+        _checkOffset(dstOffset);
+
+        if (s == null)
+            s = "null";
```
dog-m: String `length()` is known.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, b: boolean): StringBuilder
+    {
+        _checkOffset(dstOffset);
+
+        var s: String = "false";
+        if (b)
```
dog-m: String `length()` is known. In both cases.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        _insertCharSequence(dstOffset, s, len, 0, len);
+
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, c: char): StringBuilder
+    {
+        _checkOffset(dstOffset);
+        val s: String = action OBJECT_TO_STRING(c);
```
dog-m: High computational complexity. Please, optimize a bit.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, d: double): StringBuilder
+    {
+        _checkOffset(dstOffset);
+        var s: String = action OBJECT_TO_STRING(d);
+
+        if (s == null)
+            s = "null";
```
dog-m: When it can be `null`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, f: float): StringBuilder
+    {
+        _checkOffset(dstOffset);
+        var s: String = action OBJECT_TO_STRING(f);
+
+        if (s == null)
+            s = "null";
```
dog-m: When it can be `null`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, ii: int): StringBuilder
+    {
+        _checkOffset(dstOffset);
+        var s: String = action OBJECT_TO_STRING(ii);
+
+        if (s == null)
```
dog-m: When it can be `null`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = self;
+    }
+
+
+    fun *.insert (@target self: StringBuilder, dstOffset: int, l: long): StringBuilder
+    {
+        _checkOffset(dstOffset);
+        var s: String = action OBJECT_TO_STRING(l);
+
+        if (s == null)
```
dog-m: When it can be `null`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        this.storage = action OBJECT_TO_STRING(newStr);
+        this.length = newLength;
+        result = self;
+    }
+
+
+    @Phantom proc _replace_loop (i: int, strIndex: int, start: int, end: int, newStr: array<char>, str: String, strLength: int): void
+    {
+        if (i < start)
```
dog-m: I think 3 separate loops would have less computational complexity than a single one with multiple branches.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        _checkRangeSIOOBE(start, end, this.length);
+
+        val strLength: int = action CALL_METHOD(str, "length", []);
+
+        val newLength: int = this.length + strLength - (end - start);
+        var i: int = 0;
+        var strIndex: int = 0;
+        val newStr: array<char> = action ARRAY_NEW("char", newLength);
+
+        action LOOP_FOR(i, 0, newLength, +1, _replace_loop(i, strIndex, start, end, newStr, str, strLength));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    fun *.reverse (@target self: StringBuilder): StringBuilder
+    {
+        var n: int = this.length - 1;
+        var i: int = (n-1) / 2;
+        var k: int = -1;
+        val newStr: array<char> = action ARRAY_NEW("char", this.length);
+        action LOOP_FOR(i, 0, -1, -1, _reverse_loop(i, k, n, newStr));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+
+        this.storage = action OBJECT_TO_STRING(newStr);
+        result = self;
+    }
+
+
+    @Phantom proc _reverse_loop (i: int, k: int, n: int, newStr: array<char>): void
+    {
+        k = n - i;
+        newStr[k] = action CALL_METHOD(this.storage, "charAt", [i]);
```
dog-m: I don't think swaps are better than just creating a new array and copying all characters into that (also using just plus and minus one instead of *complex* math like full fledged integer subtraction).

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        newStr[i] = action CALL_METHOD(this.storage, "charAt", [k]);
+    }
+
+
+    fun *.toString (@target self: StringBuilder): String
+    {
+        action NOT_IMPLEMENTED("no final decision");
```
dog-m: Why not just return `this.storage`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        // original: return value.length >> coder;
+        // we will have another result, because we don't know "coder" value;
+        result = this.length;
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.ensureCapacity (@target self: StringBuilder, minimumCapacity: int): void
+    {
+
```
dog-m: Add a comment on why this method is empty or add `action DO_NOTHING()` to indicate the intention.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    fun *.setLength (@target self: StringBuilder, newLength: int): void
+    {
+        if (newLength < 0)
+        {
+            //val message: String = "String index out of range: " + action OBJECT_TO_STRING(newLength);
+            action THROW_NEW("java.lang.StringIndexOutOfBoundsException", []);
+        }
+        var i: int = 0;
+        val newStr: array<char> = action ARRAY_NEW("char", newLength);
+        action LOOP_FOR(i, 0, newLength, +1, _setNewLength_loop(i, newStr));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    {
+        _checkIndex(index);
+        result = action CALL_METHOD(this.storage, "charAt", [index]);
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.setLength (@target self: StringBuilder, newLength: int): void
+    {
+        if (newLength < 0)
```
dog-m: If `newLength` is exactly `0` (zero) then the following implementation is too complex.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        action LOOP_FOR(i, 0, newLength, +1, _setNewLength_loop(i, newStr));
+
+        this.storage = action OBJECT_TO_STRING(newStr);
+        this.length = newLength;
+    }
+
+
+    @Phantom proc _setNewLength_loop (i: int, newStr: array<char>): void
+    {
+        if (i < this.length)
```
dog-m: Loop bound calculations should be done before it even starts.
`else` branch is not needed: `this.storage` IS a `String` thus we do not care about trailing `null` characters.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.setCharAt (@target self: StringBuilder, index: int, ch: char): void
+    {
+        _checkIndex(index);
+        val newStr: array<char> = action ARRAY_NEW("char", this.length);
+        var i: int = 0;
+        action LOOP_FOR(i, 0, this.length, +1, _setNewLength_loop(i, index, newStr, ch));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        this.storage = action OBJECT_TO_STRING(newStr);
+    }
+
+
+    @Phantom proc _setNewLength_loop (i: int, index: int, newStr: array<char>, ch: char): void
+    {
+        if(i != index)
```
dog-m: High computational complexity. Please, optimize a bit.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+            newStr[i] = action CALL_METHOD(this.storage, "charAt", [i]);
+        else
+            newStr[i] = ch;
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.trimToSize (@target self: StringBuilder): void
+    {
+
```
dog-m: Empty function. Please add a comment with a clarification on why it is like that.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        result = _substring(start, end);
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.subSequence (@target self: StringBuilder, start: int, end: int): CharSequence
+    {
+        _checkRangeSIOOBE(start, this.length, this.length);
+        val sizeNewString: int = end - start;
+        val newStr: array<char> = action ARRAY_NEW("char", sizeNewString);
```
dog-m: What is the difference between this method and that of the above (`substring`)?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    // within java.lang.AbstractStringBuilder
+    fun *.getChars (@target self: StringBuilder, srcBegin: int, srcEnd: int, dst: array<char>, dstBegin: int): void
+    {
+        _checkRangeSIOOBE(srcBegin, srcEnd, this.length);
+        var n: int = srcEnd - srcBegin;
+        val dstLength: int = action ARRAY_SIZE(dst);
+        _checkRange(dstBegin, dstBegin + n, dstLength);
+
+        var i: int = 0;
+        action LOOP_FOR(i, srcBegin, srcEnd, +1, _getChars_loop(i, dstBegin, dst));
```
dog-m: Please, apply better formatting when loops are being used.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        action LOOP_FOR(i, start, end, +1, _newSubString_loop(i, newStr));
+        result = action OBJECT_TO_STRING(newStr);
+    }
+
+
+    // within java.lang.AbstractStringBuilder
+    fun *.getChars (@target self: StringBuilder, srcBegin: int, srcEnd: int, dst: array<char>, dstBegin: int): void
+    {
+        _checkRangeSIOOBE(srcBegin, srcEnd, this.length);
+        var n: int = srcEnd - srcBegin;
```
dog-m: Can we have a suggestion for the USVM here? What important properties we know about the value of this variable?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+        _checkRange(dstBegin, dstBegin + n, dstLength);
+
+        var i: int = 0;
+        action LOOP_FOR(i, srcBegin, srcEnd, +1, _getChars_loop(i, dstBegin, dst));
+    }
+
+
+    @Phantom proc _getChars_loop(i: int, dstBegin: int, dst: array<char>): void
+    {
+        dst[dstBegin] = action CALL_METHOD(this.storage, "charAt", [i]);
```
dog-m: Do we actually need to call `charAt` and do range checks on that? Would it be (computationally) simpler to serialize the requested segment(?) into an array of characters (using `String::toCharArray` or `String::getChars`) and then copying char-by-char?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    {
+        if (beginIndex < 0 || endIndex > this.length || beginIndex > endIndex)
+            action THROW_NEW("java.lang.IndexOutOfBoundsException", []);
+
+        val count: int = endIndex - beginIndex;
+        val newStr: array<char> = action ARRAY_NEW("char", this.length);
+        var i: int = 0;
+        action LOOP_FOR(i, 0, this.length, +1, _newSubString_loop(i, newStr));
+
+        result = action DEBUG_DO("Character.codePointCount(newStr, beginIndex, count)");
```
dog-m: This is too complex to me.
Why we can't just return an integer between [`(E - B)`...`(E - B)*2`] using `action SYBOLIC("int")`?
*(`2` - is a maximum number of code points in the largest character possible)*

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    // within java.lang.AbstractStringBuilder
+    fun *.codePointAt (@target self: StringBuilder, index: int): int
+    {
+        _checkIndex(index);
+
+        val newStr: array<char> = action ARRAY_NEW("char", this.length);
+        var i: int = 0;
+        action LOOP_FOR(i, 0, this.length, +1, _newSubString_loop(i, newStr));
+
+        result = action DEBUG_DO("Character.codePointAt(newStr, index, this.length)");
```
dog-m: Why should this be exact? Why not `SYMBOLIC`?

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    fun *.codePointBefore (@target self: StringBuilder, index: int): int
+    {
+        index -= 1;
+        _checkIndex(index);
+
+        val newStr: array<char> = action ARRAY_NEW("char", this.length);
+        var i: int = 0;
+        action LOOP_FOR(i, 0, this.length, +1, _newSubString_loop(i, newStr));
+
+        result = action DEBUG_DO("Character.codePointAt(newStr, index, this.length)");
```
dog-m: Looks too complex to me.

spec/java/util/StringBuilder.main.lsl
```diff
@@ -0,0 +1,1022 @@
...
+    // within java.lang.AbstractStringBuilder
+    fun *.chars (@target self: StringBuilder): IntStream
+    {
+        // #todo: use custom stream implementation
+        result = action SYMBOLIC("java.util.stream.IntStream");
+        action ASSUME(result != null);
+    }
+
+
+    // within java.lang.Object
```
dog-m: The original class does not have any specific implementation for `hashCode`. Thus, the following can be removed.

dog-m (changes requested):

spec/java/util/StringBuilder.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Agree with you ) Fixed;

spec/java/util/StringBuilder.main.lsl
alexeev509: Method was deleted; Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed

spec/java/util/StringBuilder.lsl
alexeev509: @dog-m If this will be deleted, method will be with another type and it willn't compile:

Original: public int compareTo(StringBuilder another)

Without '<StringBuilder>': public int compareTo(Object another)

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed in such way:

 ```
   proc _appendString (str: String): void
    {
        if (str == null) 
        {
            str = "null";
            this.length += 4;
        }
        else
        {
            this.length += action CALL_METHOD(str, "length", []);
        }
        this.storage += str;
    }
```

File: spec/java/util/StringBuilder.main.lsl (hunk 8 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 10 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: @dog-m I really don't know how it will be more simple for USVM (I don't know any details about it internal realization);

But I have understood your hint;

Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed;

File: spec/java/util/StringBuilder.main.lsl (hunk 11 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: This loop was changed and separated to three LOOP_FOR actions;

File: spec/java/util/StringBuilder.main.lsl (hunk 15 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: I thought at first - that it will be better to write in one style and "length" operation isn't difficult for "null" string for USVM;

Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 17 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 19 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 23 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 24 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 25 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 26 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 31 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 34 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 35 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 36 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 37 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 11 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 18 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: @dog-m In such way ?

```
            if (sb has StringBufferAutomaton)
            {
                this.storage += StringBufferAutomaton(sb).storage;
                this.length += StringBufferAutomaton(sb).length;
            }
```

But it doesn't work - we don't have StringBufferAutomaton at this moment; ofcourse I can describe it, but it can lead to merge conflict with Valerii;

File: spec/java/util/StringBuilder.main.lsl (hunk 20 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: @dog-m 
I have decide that for this method cycle was redudant; And replace this in such way:

    fun *.append (@target self: StringBuilder, str: array<char>): StringBuilder
    {
        val strSize: int = action ARRAY_SIZE(str);
        this.length += strSize;
        this.storage += action OBJECT_TO_STRING(str);
        result = self;
    }

File: spec/java/util/StringBuilder.lsl (hunk 2 above)
Comment (dog-m) [reply spec/java/util/StringBuilder.lsl]: Yes, it would make the Java compiler to generate the appropriate (synthetic) method for you that calls `compareTo(StringBuilder)` from `compareTo(Object)`.
If time is our priority in this case then this can be left unchanged.

File: spec/java/util/StringBuilder.main.lsl (hunk 20 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: In the second method I have made such refactoring:

```
    @Phantom proc _fromSrcArrayToDstArray_loop (i: int, arrayIndex: int, str: array<char>, subArray: array<char>): void
    {
        subArray[arrayIndex] = str[i];
    }


    fun *.append (@target self: StringBuilder, str: array<char>, offset: int, len: int): StringBuilder
    {
        val end: int = offset + len;
        val strSize: int = action ARRAY_SIZE(str);
        _checkRange(offset, end, strSize);
        val subArray: array<char> = action ARRAY_NEW("char", len);
        var arrayIndex: int = 0;
        var i: int = 0;
        action LOOP_FOR(
            i, offset, end, +1,
            _fromSrcArrayToDstArray_loop(i, arrayIndex, str, subArray)
        );
        this.length += len;
        this.storage += action OBJECT_TO_STRING(subArray);
        result = self;
    }
```

File: spec/java/util/StringBuilder.main.lsl (hunk 4 above)
Comment (dog-m) [reply spec/java/util/StringBuilder.main.lsl]: I don't see major changes here. Yes, they were moved into `StringBuilder.lsl` file but not inside of the `StringBuilder` type declaration nor got any distinct prefixes, like `STRINGBUILDER_*` or `SB_*` or something.

File: spec/java/util/StringBuilder.main.lsl (hunk 21 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 22 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 27 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 29 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Was added;

File: spec/java/util/StringBuilder.main.lsl (hunk 30 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 32 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: @dog-m It works; But only if i add declaration of indexOf to type String:

```
type String
    is java.lang.String
    for CharSequence, string
{
    fun compareTo(anotherString: string): int; // #problem: self-reference

    fun indexOf(anotherString: string): int; // #problem: self-reference
}
```

Can I change _interfaces file ?

File: spec/java/util/StringBuilder.main.lsl (hunk 33 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 39 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed;

File: spec/java/util/StringBuilder.main.lsl (hunk 40 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed;

File: spec/java/util/StringBuilder.main.lsl (hunk 41 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed;

File: spec/java/util/StringBuilder.main.lsl (hunk 42 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed;

File: spec/java/util/StringBuilder.main.lsl (hunk 38 above)
Comment (alexeev509) [reply spec/java/util/StringBuilder.main.lsl]: Fixed

File: spec/java/util/StringBuilder.main.lsl (hunk 15 above)
Comment (dog-m) [reply spec/java/util/StringBuilder.main.lsl]: Function `*.append (@target self: StringBuilder, seq: CharSequence, start: int, end: int): StringBuilder` still contains the following:
```dart
if (seq == null)
    seq = "null";
val seqLength: int = action CALL_METHOD(seq, "length", []);
```

spec/java/util/StringBuilder.main.lsl
alexeev509: Fixed;

spec/java/util/StringBuilder.main.lsl
dog-m: Why not just append `"null"` string to `this.storage` directly without char-by-char copying when it is known to be?
(taking into account the provided parameters `start` and `end`)

spec/java/util/StringBuilder.main.lsl
alexeev509: @dog-m What do in such case ? Add checking:

if (newLength == 0)... etc ?

spec/java/util/StringBuilder.main.lsl
dog-m: I suggest adding it but commented-out. I hope it will be "enabled" back into existence when `StringBuffer` be ready for merge.
Also, please add a `// #problem` comment on top of it to not forget about that.

spec/java/util/StringBuilder.main.lsl
dog-m: Please, simplify the added comment a bit.

spec/java/util/StringBuilder.main.lsl
dog-m: Action `NOT_IMPLEMENTED` is not a correct one here. Consider `DO_NOTHING` instead.

spec/java/util/StringBuilder.main.lsl
dog-m: Why not?
```dart
if (newLength < 0)
{
...
}
else if (newLength == 0)
{
...
}
else
{
...
}
```

spec/java/util/StringBuilder.main.lsl
dog-m: Action `NOT_IMPLEMENTED` is not a correct one here. Consider `DO_NOTHING` instead.

spec/java/lang/_interfaces.lsl
```diff
@@ -44,6 +44,8 @@ type String
     fun compareTo(anotherString: string): int; // #problem: self-reference
 
     fun indexOf(anotherString: string, fromIndex: int): int; // #problem: self-reference
+
+    fun toCharArray(): array<char>; // #problem: self-reference
```
dog-m: I don't see any problem here. Does it really exist or not?

spec/java/util/StringBuilder.main.lsl
dog-m: Parameter `dstOffset` is dismissed when adding a character.
I suggest creating a new array, copying characters into it from the `this.storage` *before* and *after* the destination index, and putting the provided symbol at the specified position. Then de-serializing it back into a `String` using `OBJECT_TO_STRING`.

spec/java/lang/StringBuilder.main.lsl
```diff
@@ -995,6 +1010,7 @@ automaton StringBuilderAutomaton
         _checkIndex(index);
 
         val codePoint: int = action SYMBOLIC("int");
+        action ASSUME(_isValidCodePoint(codePoint));
```
dog-m: Can we avoid using `_isValidCodePoint` here? (also below)

spec/java/util/StringBuilder.main.lsl
dog-m: I suggest the following instead:
```dart
fun *.reverse (@target self: StringBuilder): StringBuilder
{
    if (this.length != 0)
    {
        action ASSUME(this.length > 0);
        // serialize current state of storage string
        val oldStorage: array<char> = action CALL_METHOD(this.storage, "toCharArray", []);

        // prepare a new serialized but processed version of the storage
        val newStorage: array<char> = action ARRAY_NEW("char", this.length);
        action ASSUME(action ARRAY_SIZE(newStorage) == this.length);
        action ASSUME(action ARRAY_SIZE(oldStorage) == action ARRAY_SIZE(newStorage));

        var j: int = this.length - 1;
        var i: int = 0;
        action LOOP_FOR(
            i, 0, this.length, +1,
            _reverse_loop(i, oldStorage, j, newStorage)
        );

        // replace the old state with a reversed (buffer) version
        this.storage = action OBJECT_TO_STRING(newStorage);
    }
    result = self;
}

@Phantom proc _reverse_loop (i: int, oldStorage: array<char>, j: int, newStorage: array<char>): void
{
    newStorage[j] = oldStorage[i];
    j -= 1;
}
```

dog-m (approved):

PR #15, opened 2023-09-29
Improvements to actions and type declarations

dog-m:
Key changes:
- Added `GUID_NEXT` action.
- Added `IDENTITY_MAP_NEW` action.
- Added more type declarations.
- Moved type declarations for `java.lang.Object` and `java.lang.Class` into `java/lang/_interfaces`.
- Minor import clean-up.


spec/java/util/_interfaces.lsl
```diff
@@ -8,7 +8,6 @@ library std
 
 // imports
 
-import java.common;
```
alexeev509, line 11: Now this is needed to add "import java/util/function/_interfaces;" (as I understand)

In this lsl file we have Consumer;

spec/translator.actions.lsl
```diff
@@ -32,35 +31,16 @@ define action SYMBOLIC_ARRAY (
...
-        bodyProc: void   // subroutine call!
-    ): void;
-
-
-define action LOOP_BREAK (): void;
-
-
-
-/// specification development -related aspects
+define action GUID_NEXT (
```
alexeev509, line 43: I don't understand this action... Can you explain it's purpose for me ? Or write a little example ?

dog-m: Added a comment with a note on how it should behave.

alexeev509 (approved):

PR #16, opened 2023-10-04
StreamLSL and StreamLSLIterator

alexeev509:
1. Was created StreamLSL;
2. Was created StreamLSLIterator;
3. [UtStream](https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-framework/src/main/java/org/utbot/engine/overrides/stream/UtStream.java) was like sample;

Minuses:
1. Spliterator was created with "DEBUG_DO" action; This is temporary decision; (see spliterator method)
2.  Methods mapToInt, mapToLong, mapToDouble, flatMapToInt, flatMapToLong, flatMapToDouble were created with "DEBUG_DO" action; This is temporary decision; The reason of this is absence  IntStreamLSL / LongStreamLSL / DoubleStreamLSL automatons;
3. Method flatMap isn't finished because it can use Stream inside it;


spec/java/util/stream/Stream.lsl
```diff
@@ -0,0 +1,36 @@
...
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/util/stream/Stream.java";
+
+// imports
+
+import java.common;
+import java/lang/_interfaces;
+import java/util/function/_interfaces;
+import java/util/stream/_interfaces;
+import java/util/Optional;
```
dog-m: Why we need `Optional` here and not in `Stream.main` and `Stream.Iterator`?

spec/java/util/stream/Stream.lsl
```diff
@@ -0,0 +1,36 @@
...
+libsl "1.1.0";
+
+library std
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/util/stream/Stream.java";
+
+// imports
+
+import java.common;
```
dog-m: Due to `java/lang/_interfaces` already including `java.common` in its own import list, it is no longer required to be used directly. Type declaration for `Object` lies directly in `java/lang/_interfaces`, so I don't see much use of it being referenced explicitly here.

spec/java/util/stream/Stream.lsl
```diff
@@ -0,0 +1,36 @@
...
+
+// new/introduced types
+
+@GenerateMe
+@implements("java.util.stream.Stream")
+@public type StreamLSL
+    is java.util.stream.StreamLSL
+    for Stream
+{
+    @static val STREAM_VALUE: Object = 1;
```
dog-m: I suggest making it `@private` to indicate its "internal use only" nature.
Also keep in mind that this will call `Integer.valueOf` when *compiled*.

spec/java/util/stream/Stream.lsl
```diff
@@ -0,0 +1,36 @@
+//#! pragma: non-synthesizable
+libsl "1.1.0";
+
+library std
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/util/stream/Stream.java";
```
dog-m, line 7: I don't think declarations in this file belong to the specified page (others do).

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+    @Phantom proc _skip_loop (i: int, skipIndex: int, skipStorage: array<Object>): void
+    {
+        skipStorage[skipIndex] = this.storage[i];
+        skipIndex += 1;
+    }
+
+
+    fun *.forEach (@target self: Stream, _action: Consumer): void
+    {
+        _actionApply(_action);
```
dog-m, line 682: Why we need to create a new `Stream` instance and then immediately discard it here?
Same in `forEachOrdered`.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+            _closeHandlers_loop(i)
+        );
+
+        this.closeHandlers = action LIST_NEW();
+    }
+
+
+    @Phantom proc _closeHandlers_loop (i: int): void
+    {
+        action CALL(action LIST_GET(this.closeHandlers, i) as Runnable, []);
```
dog-m: A tiny bit of simplification would have been very helpful here.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        result = self;
+    }
+
+
+    // within java.lang.AutoCloseable
+    fun *.close (@target self: Stream): void
+    {
+        val listLength: int = action LIST_SIZE(this.closeHandlers);
+
+        // NOTE: this implementation does not care about suppressing and throwing exceptions produced by handlers
```
dog-m: This note originates from UtBot sources, not yourself.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        else
+        {
+            // plain bubble sorting algorithm
+            val outerLimit: int = this.length - 1;
+            var innerLimit: int = 0;
+            var i: int = 0;
+            var j: int = 0;
+            action LOOP_FOR(
+                i, 0, outerLimit, +1,
+                sort_loop_outer(i, j, innerLimit)
```
dog-m, line 448: This implementation modifies current instance but that of UtBot's does not. Is this correct?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+        val mappedStorage: array<int> = action ARRAY_NEW("int", this.length);
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
+            _mapToInt_loop(i, mapper, mappedStorage)
+        );
+
+        // Temporary decision (we don't have IntStream automaton at this moment)
```
dog-m: I suggest reformatting this comment as `// #todo: message` instead to make it easier to look up and fix it later.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+        var mappedStorage: array<long> = action ARRAY_NEW("long", this.length);
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
+            _mapToLong_loop(i, mapper, mappedStorage)
+        );
+
+        // Temporary decision (we don't have IntStream automaton at this moment)
```
dog-m: I suggest reformatting this comment as `// #todo: message` instead to make it easier to look up and fix it later.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+        var mappedStorage: array<double> = action ARRAY_NEW("double", this.length);
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
+            _mapToDouble_loop(i, mapper, mappedStorage)
+        );
+
+        // Temporary decision (we don't have IntStream automaton at this moment)
```
dog-m: I suggest reformatting this comment as `// #todo: message` instead to make it easier to look up and fix it later.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    @Phantom proc sort_loop_inner (j: int): void
+    {
+        val idxA: int = j;
+        val idxB: int = j + 1;
+        val a: Object = this.storage[idxA];
+        val b: Object = this.storage[idxB];
+
+        // "compareTo" - that's right ??
```
dog-m: Yes, the target method is `Comparator#compareTo`.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    @Phantom proc sort_loop_inner_with_comparator (j: int, comparator: Comparator): void
+    {
+        val idxA: int = j;
+        val idxB: int = j + 1;
+        val a: Object = this.storage[idxA];
+        val b: Object = this.storage[idxB];
+
+        // "compareTo" - that's right ??
```
dog-m: Target method will automatically be picket up as `compareTo` due to `Comparator` being a *functional interface*.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+                length = maxSizeInt,
+                closeHandlers = this.closeHandlers,
+            );
+        }
+    }
+
+
+    @Phantom proc _limit_loop (i: int, limitStorage: array<Object>): void
+    {
+        limitStorage[i] = this.storage[i];
```
dog-m, line 616: What if `limitStorage.length` is greater than `this.storage.length`?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+                length = this.length,
+                closeHandlers = this.closeHandlers,
+            );
+        }
+        else
+        {
+            // what will be if will be overflow ?
+            val offset: int = n as int;
+            val newLength: int = this.length - offset;
+            val skipStorage: array<Object> = action ARRAY_NEW("java.lang.Object", newLength);
```
dog-m, line 650: What if `newLength` is negative or zero?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+    fun *.forEachOrdered (@target self: Stream, _action: Consumer): void
+    {
+        _actionApply(_action);
+    }
+
+
+    fun *.toArray (@target self: Stream): array<Object>
+    {
+        result = this.storage;
```
dog-m, line 702: What would happen when user changes an element in the returned array? Is this the correct behavior?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.toArray (@target self: Stream, generator: IntFunction): array<Object>
+    {
+        val generatedArray: array<Object> = action CALL(generator,[this.length]) as array<Object>;
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
+            _copyToArray_loop(i, generatedArray)
```
dog-m: I'm pretty sure `ARRAY_COPY` action can be helpful here.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+            value = this.storage[0];
+
+            var i: int = 0;
+            action LOOP_FOR(
+                i, 1, this.length, +1,
+                _accumulate_optional_loop(i, accumulator, value)
+            );
+        }
+
+        result = action DEBUG_DO("Optional.of(value)");
```
dog-m: There will be a `NullPointerException` thrown when there are zero elements in the stream.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.reduce (@target self: Stream, identity: Object, accumulator: BinaryOperator): Object
+    {
+        if (accumulator == null || identity == null)
+            _throwNPE();
+
+        result = identity;
+
+        var i: int = 0;
```
dog-m: What if there are no elements in the stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.reduce (@target self: Stream, identity:Object, accumulator: BiFunction, combiner: BinaryOperator): Object
+    {
+        if (accumulator == null || identity == null || combiner == null)
+            _throwNPE();
+
+        result = identity;
+
+        var i: int = 0;
```
dog-m: What if there are no elements in the stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+    @Phantom proc _accumulate_optional_loop (i: int, accumulator: BinaryOperator, value: Object): void
+    {
+        value = action CALL(accumulator, [value, this.storage[i]]);
+    }
+
+
+    fun *.reduce (@target self: Stream, identity:Object, accumulator: BiFunction, combiner: BinaryOperator): Object
+    {
+        if (accumulator == null || identity == null || combiner == null)
```
dog-m: To avoid unnecessary state branching, I think it would be better to use sequential form:
```java
if (...) throw;
if (...) throw;
...
```

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+            _throwNPE();
+
+        result = identity;
```
dog-m, line 799: Why `identity` cannot be `null`?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+    @Phantom proc _accumulate_with_biFunction_loop (i: int, accumulator: BiFunction, result: Object): void
+    {
+        result = action CALL(accumulator, [result, this.storage[i]]);
+    }
+
+
+    fun *.collect (@target self: Stream, supplier: Supplier, accumulator: BiConsumer, combiner: BiConsumer): Object
+    {
+        if (supplier == null || accumulator == null || combiner == null)
```
dog-m: To avoid unnecessary state branching, I think it would be better to use sequential form.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+            _throwNPE();
+        // since this implementation is always sequential, we do not need to use the combiner
```
dog-m: This comment originates from UtBot sources. I think it should be noted explicitly.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+                i, 1, this.length, +1,
+                _find_min_loop(i, comparator, min)
+            );
+
+            result = action DEBUG_DO("Optional.of(min)");
+        }
+    }
+
+
+    @Phantom proc _find_min_loop (i: int, comparator: Comparator, min: int): void
```
dog-m: Parameter `min` is not an `int` but `Object` in the base method.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+                i, 1, this.length, +1,
+                _find_max_loop(i, comparator, max)
+            );
+
+            result = action DEBUG_DO("Optional.of(max)");
+        }
+    }
+
+
+    @Phantom proc _find_max_loop (i: int, comparator: Comparator, max: int): void
```
dog-m: Parameter `max` is not an `int` but `Object` in the base method.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        );
+    }
+
+
+    @Phantom proc _anyMatch_loop (i: int, predicate: Predicate, result: boolean): void
+    {
+        if (action CALL(predicate, [this.storage[i]]))
+        {
+            result = true;
+            action LOOP_BREAK();
```
dog-m: A `LOOP_WHILE` can be used here instead of calling `LOOP_BREAK`.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        );
+    }
+
+
+    @Phantom proc _allMatch_loop (i: int, predicate: Predicate, result: boolean): void
+    {
+        if (!action CALL(predicate, [this.storage[i]]))
+        {
+            result = false;
+            action LOOP_BREAK();
```
dog-m: A `LOOP_WHILE` can be used here instead of calling `LOOP_BREAK`.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.anyMatch (@target self: Stream, predicate: Predicate): boolean
+    {
+        if (predicate == null)
+            _throwNPE();
+
+        result = false;
+
+        var i: int = 0;
```
dog-m, line 966: What if there no elements in this stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.allMatch (@target self: Stream, predicate: Predicate): boolean
+    {
+        if (predicate == null)
+            _throwNPE();
+
+        result = true;
+
+        var i: int = 0;
```
dog-m: What if there no elements in this stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    fun *.noneMatch (@target self: Stream, predicate: Predicate): boolean
+    {
+        if (predicate == null)
+            _throwNPE();
+
+        result = true;
+
+        var i: int = 0;
```
dog-m: What if there no elements in this stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        );
+    }
+
+
+    @Phantom proc _noneMatch_loop (i: int, predicate: Predicate, result: boolean): void
+    {
+        if (action CALL(predicate, [this.storage[i]]))
+        {
+            result = false;
+            action LOOP_BREAK();
```
dog-m: A `LOOP_WHILE` can be used here instead of calling `LOOP_BREAK`.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+        val listLength: int = action LIST_SIZE(this.closeHandlers);
+
+        // NOTE: this implementation does not care about suppressing and throwing exceptions produced by handlers
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, listLength, +1,
+            _closeHandlers_loop(i)
+        );
+
+        this.closeHandlers = action LIST_NEW();
```
dog-m, line 1140: Is this possible to use (call other methods, like `min/max`) this stream after closing it?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+    }
+
+
+    @Phantom proc _closeHandlers_loop (i: int): void
+    {
+        action CALL(action LIST_GET(this.closeHandlers, i) as Runnable, []);
+    }
+
+
+    // Do I need @default annotation for this method ?
```
dog-m: Annotation `@default` is useful only for interfaces.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+    fun *.dropWhile (@target self: Stream, predicate: Predicate): Stream
+    {
+        if (predicate == null)
+            _throwNPE();
+
+        var dropLength: int = 0;
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
```
dog-m: What if there are no elements in this stream?

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+
+
+    @Phantom proc _copy_dropWhile_loop (i: int, j: int, newStorage: array<Object>): void
+    {
+        newStorage[j] = this.storage[i];
+        j += 1;
+    }
+
+
+    // Do I need @default annotation for this method ?
```
dog-m: Annotation `@default` is useful only for interfaces.

spec/java/util/stream/Stream.main.lsl
```diff
@@ -0,0 +1,1049 @@
...
+    fun *.takeWhile (@target self: Stream, predicate: Predicate): Stream
+    {
+        if (predicate == null)
+            _throwNPE();
+
+        var takeLength: int = 0;
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, this.length, +1,
```
dog-m: What if there are no elements in this stream?

spec/java/util/stream/Stream.Iterator.lsl
```diff
@@ -0,0 +1,148 @@
...
+
+
+    // methods
+
+    fun *.hasNext (@target self: StreamLSLIterator): boolean
+    {
+        // relax state/error discovery process
+        action ASSUME(this.parent != null);
+
+        result = this.cursor != StreamAutomaton(this.parent).length;
```
dog-m, line 41: Can value of `length` be changed during lifetime of this iterator?

spec/java/util/stream/Stream.Iterator.lsl
```diff
@@ -0,0 +1,148 @@
...
+        if (i >= action ARRAY_SIZE(parentStorage))
+            _throwCME();
+
+        this.cursor = i + 1;
+        this.lastRet = i;
+        result = parentStorage[i];
+    }
+
+
+    fun *.remove (@target self: StreamLSLIterator): void
```
dog-m, line 61: I don't see support for element removal in UtBot sources. Is this correct?

spec/java/util/stream/Stream.Iterator.lsl
```diff
@@ -0,0 +1,148 @@
...
+        // relax state/error discovery process
+        action ASSUME(this.parent != null);
+
+        if (userAction == null)
+            action THROW_NEW("java.lang.NullPointerException", []);
+
+        var i: int = this.cursor;
+        val size: int = StreamAutomaton(this.parent).length;
+
+        if (i < size)
```
dog-m: Why not use `!=` operator instead?

spec/java/util/stream/Stream.Iterator.lsl
```diff
@@ -0,0 +1,148 @@
...
+        {
+            val pStorage: array<Object> = StreamAutomaton(this.parent).storage;
+
+            if (i >= action ARRAY_SIZE(pStorage))
+                _throwCME();
```
dog-m: Can this happen?

dog-m (changes requested):

spec/java/util/stream/Stream.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.lsl
alexeev509: Fixed (it was deleted because it was redundant )

spec/java/util/stream/Stream.lsl
alexeev509: Fixed (was added @private)

spec/java/util/stream/Stream.lsl
alexeev509: @dog-m It must be empty ?

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m Yes, but as I see this comment is needed here; Was removed "throws exception";

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m As I see it isn't principle - to use storage (or copy of the storage); If I'm not right - please write example for me )

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m In original method willn't be any exception; It will return all elements;

```
        Stream<Integer> integerStream1 = Stream.of(5, 3, 6);

        integerStream1.limit(7).forEach(System.out::println);
```

alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
dog-m: I've meant that it would be better to format this comment something like this: `// UtBot note: text`.

spec/java/util/stream/Stream.main.lsl
dog-m: Yes if the stream is handled correctly by the end user.

Interestingly, this implementation (and that of UtBot) does not capture incorrect usage of streams as I can understand it. Consider the following code:
```java
var s = new Random(123).ints(5).limit(20);
var a = s.filter(i -> i % 2 == 0);
var b = s.sorted();  // !!!
```

The original `Stream#sorted` throws an exception with a message *"IllegalStateException: stream has already been operated upon or closed"*.
Should this implementation be improved to also detect illegal usages like that or not?

spec/java/util/stream/Stream.main.lsl
dog-m: Streams are generally a one-time-use objects. How it would be modeled here?
Consider the following use case:
```java
var s = new Random(123).ints(5).limit(20);
var a = s.toArray();
var b = s.toArray();  // !!!
```
The original `Stream` would throw an exception: `IllegalStateException: stream has already been operated upon or closed`.

spec/java/util/stream/Stream.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m 
In such example result == identity;

```
        var s11 = Stream.of().reduce(5, (x, y) -> y);
        System.out.println(s11);
```
I have mistake ?

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m 
In such example result == identity;

```
        var s11 = Stream.of().reduce(5, (x, y) -> y);
        System.out.println(s11);
```
I have mistake ?

spec/java/util/stream/Stream.main.lsl
alexeev509: I fixed; 

But it really will be better for USVM ?

spec/java/util/stream/Stream.main.lsl
alexeev509: fixed;

spec/java/util/stream/Stream.main.lsl
alexeev509: fixed;

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed;

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m No - it will be thrown exception "stream has already been operated upon or closed";
What I can do with this ? Add boolean flag "closed" to StreamAutomaton ?

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed (I suppose)

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed (I suppose)

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m Maybe add flag "arrayWasCreated" (for example) ? If it was true - throw IllegalStateException ? 
What do you think ?

spec/java/util/stream/Stream.main.lsl
alexeev509: @dog-m I think it depends on approximation accuracy... I can't answer how accurate we must do this approximation; 
I can add flag... Add this flag or not ? What do you think ? 

Maybe ask Valentin about this ?

spec/java/util/stream/Stream.Iterator.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.Iterator.lsl
alexeev509: Why do you think that this will be better ? (I really don't understand)

spec/java/util/stream/Stream.Iterator.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
dog-m: `var i: int = 0;` left hanging around.

spec/java/util/stream/Stream.main.lsl
dog-m: I suggest adding a `var closed: boolean = false;` internal automaton variable and **updating and checking** it in every(?) *chaining* (that what allows chaining methods together) and *terminal* (that what returns a final processed value: an array, a value, etc.) method.

spec/java/util/stream/Stream.main.lsl
dog-m: But what if the received `accumulator` always return `null`? That will also cause `Optional#of` method to throw an exception.
Why not use `Optional#ofNullable` instead?

Same for `_findFirst`, `min` and `max`.

spec/java/util/stream/Stream.main.lsl
dog-m: I think this would be a bit better:
```dart
if (this.length != 0)
{
    action ASSUME(this.length > 0);
...
```

spec/java/util/stream/Stream.main.lsl
dog-m: I agree with adding a `var closed: boolean = false;` internal automaton variable and **updating and checking** it in every(?) *chaining* (that what allows chaining methods together) and *terminal* (that what returns a final processed value: an array, a value, etc.) method.

spec/java/util/stream/Stream.main.lsl
dog-m: Please avoid using the same `@Phantom proc` as loop body for different instances of `LOOP_*` actions. This makes it confusing to read and follow through the control logic and data flow.

I suggest the following style:
```dart
fun foo (...): ...
{
    action LOOP_WHILE(
        result && a < b,
        foo_loop(a, b, result)   // if there are only a single loop in this function/subroutine
        // foo_something (...)   // if there is multiple loops, named with distinct sounding name
        // foo_loop_outer (...)  // if there is nested loops
    );
}

@Phantom proc foo_loop (...): void
{
...
}
```

spec/java/util/stream/Stream.main.lsl
dog-m: Please avoid using the same `@Phantom proc` as loop body for different instances of `LOOP_*` actions. This makes it confusing to read and follow through the control logic and data flow.

spec/java/util/stream/Stream.main.lsl
dog-m: Please avoid using the same `@Phantom proc` as loop body for different instances of `LOOP_*` actions. This makes it confusing to read and follow through the control logic and data flow.

spec/java/util/stream/Stream.main.lsl
dog-m: I agree with using `closed` flag idea.

spec/java/util/stream/Stream.Iterator.lsl
dog-m: Decided to leave it as is for now.

spec/java/util/stream/Stream.Iterator.lsl
dog-m: As I can understand it, the "less-than" operator implies that the result (of such operation) is `true` when the difference between operands is negative (keeping in mind that we are dealing with *signed* types here): `(a < b) == (a - b <' 0) == isNeg(a - b)`. The "sign" of the value is usually stored as a `0` or `1` at the specific location in memory/on stack near the magnitude of the value (in some form). Thus `a < b` is essentially a (multi-bit) subtraction and a (single) bit check.

The "(not)equals" operator on the other hand checks equivalence of all bits at the same locations:
```
(a != b) <->
(a_0 !=' b_0 || a_1 !=' b_1 || ... || a_31 !=' b_31) <->
(a xor b !=' 0) <->
(a_0 xor b_0 || a_1 xor b_1 || ... || a_31 xor b_31)
```
Thus `a != b` is essentially a (multi-bit) "xor" with (complex) "or".

That said, subtraction, in my eye, is much more computationally expensive than "xor" and "*complex* or" combined (I'm not talking about "xor" in general right now).

And, yes, it is highly dependent on the capabilities of the solver that is being used: does it have optimized implementations for `<` and `!=` operators or not.
As I can remember, Valentin told us during first offline meeting, that it is preferable to use equals operator over other ones. But I might have missed the point or misunderstood something ¯\\(°_o)/¯.

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

spec/java/util/stream/Stream.main.lsl
alexeev509: Fixed

dog-m (approved):

PR #17, opened 2023-10-09
Spec for `java.util.LinkedHashSet`

chernyiv:
Added specification for LinkedHashSet


.idea/jsl-spec.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dog-m: There should be no *user-specific* IDE project files.

dog-m (changes requested):

dog-m (approved):

PR #21, opened 2023-11-13
Specs for `java.lang.System`, `java.lang.Integer`, `java.lang.Float`, `java.util.concurrent.atomic.AtomicBoolean/Integer/Long/Reference` + Minor improvements to `java.util.ArrayList`, `java.util.LinkedList`

dog-m:
Key changes:
- Added spec for `java.lang.Float`.
- Added spec for `java.lang.System`.
- Added spec for `java.lang.SecurityManager`.
- Added spec for `java.util.concurrent.atomic.AtomicBoolean`.
- Added spec for `java.util.concurrent.atomic.AtomicInteger`.
- Added spec for `java.util.concurrent.atomic.AtomicLong`.
- Added spec for `java.util.concurrent.atomic.AtomicReference`.
- Added implementation for some incomplete methods in `java.util.ArrayList` (and relevant classes).
- Added implementation for some incomplete methods in `java.util.LinkedList` (and relevant classes).
- Added partial implementation for `java.util.Spliterators`.
- Added more (relevant) type declarations.
- Repurposed `java.util.stream.*Spliterator` classes for use as companions of `java.util.Spliterators`.
- Introduced utility classes:
   * `runtime/utils/SymbolicInputStream`
   * `runtime/utils/VoidInputStream`
   * `runtime/utils/VoidOutputStream`
- Minor fixes for `java.lang.StringBuffer`.
- Using a newer naming convention for special methods (eg. `<clinit>` instead of `__clinit__`).
- Skeletons for future directions.


spec/java/io/Console.lsl
```diff
@@ -0,0 +1,27 @@
...
+// primary semantic types
+
+@final type Console
+    is java.io.Console
+    for Flushable
+{
+}
+
+// see java.io.Console#istty
+val CONSOLE_IS_TTY: boolean = action SYMBOLIC("boolean");
```
alexeev509, line 23: What do this variable ? 
I have looked at Console class but didn't understand;

spec/java/lang/ClassLoader.lsl
```diff
@@ -0,0 +1,26 @@
...
+import java/io/Serializable;
+import java/lang/String;
+
+
+// primary semantic types
+
+@final type ClassLoader
+    is java.lang.ClassLoader
+    // #problem: no reflection support
+    for Serializable //, GenericDeclaration, Type, AnnotatedElement
```
alexeev509: Why here is "Serializable" ?

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+library std
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/lang/Double.java";
+
+// imports
+
+import java/lang/Object;
+import java/lang/String;
+
```
alexeev509, line 12: Do we need this space ? Or it was added specially ? In previous specs we didn't add this;

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+        action TODO();
+    }
+
+
+    // constructors
+
+    @throws(["java.lang.NumberFormatException"])
+    @Phantom constructor *.Double (@target self: LSLDouble, s: String)
+    {
+        // NOTE: using original method
```
alexeev509, line 120: Are we using original method because there is very comprehensive implementation in original code ?

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+
+    // internal variables
+
+    // utilities
+
+    @static proc _getRawBits (v: double): long
+    {
+        if (v != v) // a NaN?
+            result = 9221120237041090560L;
+        else if (1.0 / v == NEGATIVE_INFINITY) // is it a "-0.0" ?
```
alexeev509, line 77: Can you explain this case, please ? I didn't understand why it was described here too...

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+
+    @static fun *.hashCode (value: double): int
+    {
+        result = _getRawBits(value) as int;
+    }
+
+
+    @static fun *.isFinite (d: double): boolean
```
alexeev509, line 167: Do we need to add NaN case ? 
Result of this code is false;
`System.out.println(Double.isFinite(Double.NaN));`

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+            action ASSUME(result == result);
+            action ASSUME(result != POSITIVE_INFINITY);
+            action ASSUME(result != NEGATIVE_INFINITY);
+        }
+    }
+
+
+    @static fun *.max (a: double, b: double): double
+    {
+        if (a > b)
```
alexeev509: Do we need to add such code like in original class, or not ? 

```
        if (a != a)
            return a;   // a is NaN
```

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+            result = a;
+        else
+            result = b;
+    }
+
+
+    @static fun *.min (a: double, b: double): double
+    {
+        if (a < b)
```
alexeev509: Do we need to add such code like in original class, or not ? 

```
        if (a != a)
            return a;   // a is NaN
```

spec/java/lang/Double.main.lsl
```diff
@@ -0,0 +1,375 @@
...
+    }
+
+
+    fun *.compareTo (@target self: LSLDouble, anotherDouble: Double): int
+    {
+        val a: double = this.value;
+        val b: double = DoubleAutomaton(anotherDouble).value;
+
+        // #problem: does not catch (-0.0, 0.0)
+        if (a == b || a != a || b != b) // include NaN's
```
alexeev509, line 319: Maybe add "proc"  for compareTo and compare ?

alexeev509: Or it must be static and this is bad ?

spec/java/lang/Float.lsl
```diff
@@ -24,3 +25,31 @@ import java/lang/Number;
 
 // global aliases and type overrides
 
+// note: this is a partial implementation, no need for additional constraints (abstract class) and synthetic methods (Comparable)
+// @extends("java.lang.Number")
```
alexeev509: Why this inheritance is commented everywhere ?

spec/java/lang/Float.lsl
```diff
@@ -24,3 +25,31 @@ import java/lang/Number;
...
+    @public @static val MIN_VALUE: float = action DEBUG_DO("1.4e-45f");
+
+    @public @static val MIN_NORMAL: float = action DEBUG_DO("1.17549435e-38f");
+
+    @public @static val NEGATIVE_INFINITY: float = -1.0f / 0.0f;
+    @public @static val POSITIVE_INFINITY: float = 1.0f / 0.0f;
+    @public @static val NaN: float = 0.0f / 0.0f;
+
+    // #problem: unable to get reference to primitive type
+    @public @static val TYPE: Class = action DEBUG_DO("Float.class"); // preventing recursion
```
alexeev509: Where this field is used ? Or in another automatons (not FloatAutomaton) ?

alexeev509: Maybe "java.lang.Float.class" ? Like in Integer below ?

spec/java/lang/Integer.main.lsl
```diff
@@ -0,0 +1,522 @@
...
+        }
+    }
+
+
+    @static fun *.compareUnsigned (x: int, y: int): int
+    {
+        x += MIN_VALUE;
+        y += MIN_VALUE;
+
+        if (x == y)
```
alexeev509, line 154: Maybe add this lines to static proc ?

spec/java/lang/Integer.main.lsl
```diff
@@ -0,0 +1,522 @@
...
+            // direct adaptation from the JDK
+            var y: int = 0;
+            var n: int = 31;
+
+            y = i << 16;  if (y != 0) { n -= 16; i = y; }
+            y = i <<  8;  if (y != 0) { n -=  8; i = y; }
+            y = i <<  4;  if (y != 0) { n -=  4; i = y; }
+            y = i <<  2;  if (y != 0) { n -=  2; i = y; }
+
+            result = n - ((i << 1) >>> 31);
```
alexeev509: Maybe use not "n" and only "result" ? In one style like in "numberOfLeadingZeros"; Or this is impossible ?

spec/java/lang/Integer.main.lsl
```diff
@@ -0,0 +1,522 @@
...
+
+    fun *.byteValue (@target self: LSLInteger): byte
+    {
+        result = this.value as byte;
+    }
+
+
+    @Phantom fun *.compareTo (@target self: LSLInteger, anotherInteger: Integer): int
+    {
+        // NOTE: using the original method
```
alexeev509: Why this method doesn't have body ? (As I see this is plain to describe; But maybe I don't see problems with spec for this)

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+                perm
+            ]);
+    }
+
+
+    // constructors
+
+    constructor *.LSLSecurityManager (@target self: LSLSecurityManager)
+    {
+        val actionName: String = "createSecurityManager";
```
alexeev509: Maybe "permissionName" ?

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+            action THROW_NEW("java.security.AccessControlException", [
+                "access denied " /* + action OBJECT_TO_STRING(perm) */,
+                perm
+            ]);
+    }
+
+
+    // constructors
+
+    constructor *.LSLSecurityManager (@target self: LSLSecurityManager)
```
alexeev509: Code in this constructor must be synchronized (or not) ? 


```
    public SecurityManager() {
        synchronized(SecurityManager.class) {
            SecurityManager sm = System.getSecurityManager();
            if (sm != null) {
                // ask the currently installed security manager if we
                // can create a new one.
                sm.checkPermission(new RuntimePermission
                                   ("createSecurityManager"));
            }
            initialized = true;
        }
    }
```

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+
+    @AutoInline @Phantom proc _throwIAE (): void
+    {
+        action THROW_NEW("java.lang.IllegalArgumentException", []);
+    }
+
+
+    proc _do_checkPermission (perm: Permission): void
+    {
+        if (action SYMBOLIC("boolean"))
```
alexeev509, line 89: Maybe !(action SYMBOLIC("boolean")) ? Or not ? (I could be wrong here)

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+        _do_checkPermission(
+            null // #todo
+        );
+    }
+
+
+    fun *.checkAccess (@target self: LSLSecurityManager, t: Thread): void
+    {
+        if (t == null)
+            action THROW_NEW("java.lang.NullPointerException", ["thread can't be null"]);
```
alexeev509, line 129: Do we need to add text to exception ? As I remember usvm had problems with it... ?

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+        _do_checkPermission(
+            null // #todo
+        );
+    }
+
+
+    fun *.checkAccess (@target self: LSLSecurityManager, g: ThreadGroup): void
+    {
+        if (g == null)
+            action THROW_NEW("java.lang.NullPointerException", ["thread group can't be null"]);
```
alexeev509, line 142: Do we need to add text to exception ? As I remember usvm had problems with it... ?

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+        _do_checkPermission(
+            null // #todo
+        );
+    }
+
+
+    fun *.checkConnect (@target self: LSLSecurityManager, host: String, port: int): void
+    {
+        if (host == null)
+            action THROW_NEW("java.lang.NullPointerException", ["host can't be null"]);
```
alexeev509, line 155: Do we need to add text to exception ? As I remember usvm had problems with it... ?

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+        _do_checkPermission(
+            null // #todo
+        );
+    }
+
+
+    fun *.checkConnect (@target self: LSLSecurityManager, host: String, port: int, context: Object): void
+    {
+        if (host == null)
+            action THROW_NEW("java.lang.NullPointerException", ["host can't be null"]);
```
alexeev509, line 170: Do we need to add text to exception ? As I remember usvm had problems with it... ?

spec/java/lang/SecurityManager.main.lsl
```diff
@@ -0,0 +1,411 @@
...
+    fun *.getSecurityContext (@target self: LSLSecurityManager): Object
+    {
+        result = action SYMBOLIC("java.security.AccessControlContext");
+        action ASSUME(result != null);
+    }
+
+
+    @Phantom fun *.getThreadGroup (@target self: LSLSecurityManager): ThreadGroup
+    {
+        // NOTE: using the original method
```
alexeev509: Why is original method here ? It invokes native method...
```
    public ThreadGroup getThreadGroup() {
        return Thread.currentThread().getThreadGroup();
    }
```
`    public static native Thread currentThread();`
Maybe it must be TODO ? Or this is impossible to describe ?

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+
+    // constructors
+
+    // static methods
+
+    // methods
+
+    fun *.append (@target self: System_PrintStream, csq: CharSequence): PrintStream
+    {
+        if (csq == null)
```
alexeev509: I dont see throwNPE         

```
if (csq == null)
            print("null");
```

alexeev509: [csq - The character sequence to append. If csq is null, then the four characters "null" are appended to this output stream.](https://docs.oracle.com/javase/8/docs/api/java/io/PrintStream.html#append-java.lang.CharSequence-)

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+            _throwNPE();
+
+        _checkOpen();
+
+        result = self as Object as PrintStream;
```
alexeev509, line 97: Why it can't be "result = self" ? 

I remember your idea from our dialog:

> Но это привело к тому, что при создании экземпляра объекта вызывается супер-конструктор: super(...)
> И без этого никуда - javac не даст собрать, так как это обязательная фича языка Java, - инициализация родителя.
> 
> И мне буквально на днях пришла идея: а ведь всё наследуется от Object, и при этом мы можем, если нам передан объект такого типа, уточнить (кастом), что это вот, например, PrintStream:

But we don't create object here; We return same object

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+
+    fun *.append (@target self: System_PrintStream, csq: CharSequence, start: int, end: int): PrintStream
+    {
+        if (csq == null)
```
alexeev509, line 103: I dont see throwNPE

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+
+        _checkOpen();
+
+        result = self as Object as PrintStream;
+    }
+
+
+    fun *.append (@target self: System_PrintStream, c: char): PrintStream
+    {
+        _checkOpen();
```
alexeev509, line 118: I don't understand this approximation... Can you explain please ? 

And How we can approximate "writing to the output  stream" ?

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+
+    fun *.checkError (@target self: System_PrintStream): boolean
+    {
+        result = this.error;
+    }
+
+
+    fun *.close (@target self: System_PrintStream): void
+    {
+        this.closed = true;
```
alexeev509, line 133: In original method this block is synchronized

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+
+    fun *.flush (@target self: System_PrintStream): void
+    {
+        _checkOpen();
```
alexeev509, line 140: In original method this block is synchronized

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+    }
+
+
+    @varargs fun *.format (@target self: System_PrintStream, l: Locale, format: String, args: array<Object>): PrintStream
+    {
+        if (l == null || format == null || args == null)
```
alexeev509: As I see in description of this method "NullPointerException – If the format is null";

Maybe you check another fields ("format == null || args == null") because they can throw NPE not in method "format" but inside another methods which invoke inside "format" ?

spec/java/lang/System.StdOut.lsl
```diff
@@ -0,0 +1,348 @@
...
+
+    fun *.println (@target self: System_PrintStream, x: long): void
+    {
+        _checkOpen();
+    }
+
+
+    @throws(["java.io.IOException"])
+    // within java.io.OutputStream
+    fun *.write (@target self: System_PrintStream, b: array<byte>): void
```
alexeev509, line 310: Sorry, but I can't find this method in JDK 1.8...

spec/java/lang/System.main.lsl
```diff
@@ -0,0 +1,566 @@
...
+    {
+        // doing nothing - this is a (singleton) utility class
+    }
+
+
+    // static methods
+
+    @Phantom @static fun *.arraycopy (src: Object, srcPos: int, dest: Object, destPos: int, length: int): void
+    {
+        // WARNING: do not approximate this method - infinite recursion!
```
alexeev509, line 235: This is really impossible to approximate this method ?

spec/java/lang/System.main.lsl
```diff
@@ -0,0 +1,566 @@
...
+        if (action MAP_HAS_KEY(propsMap, key))
+            result = action MAP_GET(propsMap, key);
+        else
+            result = def;
+    }
+
+
+    @Phantom @static fun *.getSecurityManager (): SecurityManager
+    {
+        // NOTE: using the original method
```
alexeev509: As I see we can set SecurityManager with method "setSecurityManager"; Why we can't return this ? Or approximation of this method doesn't have any sense ?

spec/java/lang/System.main.lsl
```diff
@@ -0,0 +1,566 @@
...
+    }
+
+
+    @static fun *.load (filename: String): void
+    {
+        if (filename == null)
+            _throwNPE();
+
+        // faking underlying checks and loading procedures
+        if (action SYMBOLIC("boolean")) action THROW_NEW("java.lang.SecurityException",    ["<message>"]);
```
alexeev509, line 420: Do we need to add message for exceptions ? 
Previously (as I remember) we deleted messages everywhere for exceptions...

spec/java/lang/System.main.lsl
```diff
@@ -0,0 +1,566 @@
...
+    {
+        if (stream == null)
+            _throwNPE();
+
+        // #todo: add checks and exceptions
+        in = stream;
+    }
+
+
+    @static fun *.setOut (stream: PrintStream): void
```
alexeev509, line 482: I had run such code: "System.setOut(null);"; I didn't throw NPE;

spec/java/util/ArrayList.SubList.ListIterator.lsl
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+
+    fun *.forEachRemaining (@target self: ArrayList_SubList_ListIterator, userAction: Consumer): void
+    {
+        // relax state/error discovery process
+        action ASSUME(this.root != null);
+
+        if (userAction == null)
+            action THROW_NEW("java.lang.NullPointerException", []);
```
alexeev509, line 215: Maybe _throwNPE() ? (Specs in one style maybe more preferable; I understand that this is one place for this exception)

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
```diff
@@ -0,0 +1,337 @@
...
+            else
+                this.value = FALSE;
+        }
+    }
+
+
+    @final fun *.compareAndSet (@target self: LSLAtomicBoolean, expectedValue: boolean, newValue: boolean): boolean
+    {
+        val currentValue: boolean = this.value != FALSE;
+        if (currentValue == expectedValue)
```
alexeev509, line 126: Can we check in such way: "this.value == expectedValue" ? Or this is incorrect ?

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
```diff
@@ -0,0 +1,337 @@
...
+        // #problem: unable to model memory effects here
+        result = this.value != FALSE;
+    }
+
+
+    @final fun *.getAndSet (@target self: LSLAtomicBoolean, newValue: boolean): boolean
+    {
+        result = this.value != FALSE;
+
+        if (newValue)
```
alexeev509, line 160: Do we need this checking ? MAYBE: this.value = newValue;  ?

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
```diff
@@ -0,0 +1,337 @@
...
+
+
+    @final fun *.compareAndSet (@target self: LSLAtomicBoolean, expectedValue: boolean, newValue: boolean): boolean
+    {
+        val currentValue: boolean = this.value != FALSE;
+        if (currentValue == expectedValue)
+        {
+            result = true;
+
+            if (newValue)
```
alexeev509, line 130: Do we need this checking ? MAYBE: this.value = newValue; ?

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
```diff
@@ -0,0 +1,337 @@
...
+    @final fun *.getPlain (@target self: LSLAtomicBoolean): boolean
+    {
+        result = this.value != FALSE;
+    }
+
+
+    @final fun *.lazySet (@target self: LSLAtomicBoolean, newValue: boolean): void
+    {
+        // #problem: delayed assignment in multithreaded environment
+        if (newValue)
```
alexeev509, line 183: Do we need this checking ? MAYBE: this.value = newValue; ?

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
```diff
@@ -0,0 +1,337 @@
...
+        if (this.value == FALSE)
+            result = "false";
+        else
+            result = "true";
+    }
+
+
+    fun *.weakCompareAndSet (@target self: LSLAtomicBoolean, expectedValue: boolean, newValue: boolean): boolean
+    {
+        // #problem: delayed assignment in multithreaded environment
```
alexeev509, line 240: Maybe another reason of the problem ? For example: 
"// #problem: This assignment may fail spuriously (typically, due to memory contention); We can't control this;"

spec/java/util/stream/DoubleStream.lsl
```diff
@@ -30,6 +30,10 @@ import java/util/Spliterator;
     is java.util.stream.DoubleStreamLSL
     for DoubleStream
 {
+    @private @static val DOUBLE_POSITIVE_INFINITY: double = 1.0 / 0.0;
```
alexeev509, line 33: Why don't you want make this constants global ? For Double / DoubleStream and maybe another classes ?

spec/java/util/stream/DoubleStream.lsl
```diff
@@ -45,7 +49,7 @@ import java/util/Spliterator;
 @GenerateMe
 @implements("java.util.Spliterator.OfDouble")
 @public type DoubleStreamLSLSpliterator
-    is java.util.DoubleStreamLSLSpliterator
+    is java.util.stream.DoubleStreamLSLSpliterator
```
alexeev509: I suppose this line has sense... But I thought that it will be prototype of "DoubleArraySpliterator" and must be in same package ("java.util");

spec/runtime/utils/SymbolicInputStream.lsl
```diff
@@ -0,0 +1,21 @@
...
+
+import java/io/InputStream;
+
+
+// primary semantic types
+
+@GenerateMe
+@extends("java.io.InputStream")
+@public @final type SymbolicInputStream
+    is runtime.utils.SymbolicInputStream
```
alexeev509: the prototype of which class is this class ? (I can't find such class "runtime.utils.SymbolicInputStream")

spec/runtime/utils/VoidInputStream.lsl
```diff
@@ -0,0 +1,22 @@
...
+// imports
+
+import java/io/InputStream;
+
+
+// primary semantic types
+
+@GenerateMe
+@extends("java.io.InputStream")
+@public @final type VoidInputStream
```
alexeev509: Why does this name have prefix "Void" ?

spec/runtime/utils/VoidOutputStream.main.lsl
```diff
@@ -0,0 +1,99 @@
...
+    {
+        // does nothing just like the original
+    }
+
+
+    @throws(["java.io.IOException"])
+    // within java.io.OutputStream
+    fun *.write (@target self: VoidOutputStream, b: array<byte>): void
+    {
+        if (b == null)
```
alexeev509: Do we need npe here ?

spec/java/io/Console.lsl
dog-m: As I understand, the referred method should tell if a Java process is run using an "human-interactive" shell or sending bytes in and out some files that model input and output buffers.
This information can be used to display colored text or any other fancy elements like text-based progress bars, tables, "windows" and other TUI elements.

spec/java/lang/Double.main.lsl
dog-m: The blank line was added on purpose to visually separate "*just dependencies*" from the actual class that this model is for.

spec/java/lang/Double.main.lsl
dog-m: This is the only way I can think of to make it possible to check for *negative zeros* since we do not have direct bit-level access to underlying structures of `float`s and `double`s.

spec/java/lang/Double.main.lsl
dog-m: It is too complex for me. We will add a proper (and probably simpler than the original) implementation when this become an issue.

spec/java/lang/Double.main.lsl
dog-m: I would prefer to not use additional calls within base classes too much. Thus it is copy-pasted.

spec/java/lang/SecurityManager.main.lsl
dog-m: Given:
```
symbolic(boolean) -> {false, true}
not symbolic(boolean) -> {false, true}
```
What is the difference?
As I know, we do not have control over the order in which the symbolic executor will choose to look through the allowed value options.

spec/java/lang/SecurityManager.main.lsl
dog-m: Static string literals do not have such problems, if I'm not mistaken.

spec/java/lang/SecurityManager.main.lsl
dog-m: Static string literals do not have such problems, if I'm not mistaken.

spec/java/lang/SecurityManager.main.lsl
dog-m: Static string literals do not have such problems, if I'm not mistaken.

spec/java/lang/SecurityManager.main.lsl
dog-m: Static string literals do not have such problems, if I'm not mistaken.

spec/java/lang/SecurityManager.main.lsl
dog-m: Modelling `Thread` class is a completely separate (an complex) task on its own.
This is why I've decided to left this like it currently is.

spec/java/lang/System.StdOut.lsl
dog-m: I'm not sure on how the current implementation of method dispatching works in USVM.
As I can see from `org.gradle.internal.io.LinePerThreadBufferingOutputStream` class (as the only `@Override`ing example I have) its implementation for method `append(java.lang.CharSequence)` return a reference to an object that is a sub-class of `java.io.PrintStream`, when the translated (using the LibSL translator tool) *implementation* class (i.e. `generated.java.lang.System_PrintStream`) is NOT.

The latter is important since we do not want to trigger `init` of `java.io.OutputStreamWriter` because of [Charset issues](https://github.com/UnitTestBot/java-stdlib-approximations/issues/3).

spec/java/lang/System.StdOut.lsl
dog-m: This overall implementation assumes that everything is being written into nothingness, never being received back.
So it is effectively the same to do nothing at all with the data being send out.

And, yes, it is *technically* possible to construct a program (or rather, a pair of programs) that writes a piece of text into the `stdout` and then reads it back via `stdin` expecting it to be exactly the same and crashing otherwise.

To (partially) eliminate such cases, `System.in` is being modeled via `SymbolicInputStream` which is capable of generating arbitrary strings of data, making it possible to explore both cases: *"the same"* and *"not the same"*.

spec/java/lang/System.StdOut.lsl
dog-m: It makes sense only in the original `java.io.PrintStream` because of its complexity.
Also, parallel execution is not being modeled nor supported.

spec/java/lang/System.StdOut.lsl
dog-m: I does not make any difference in single-threaded model.

spec/java/lang/System.StdOut.lsl
dog-m: [There](https://github.com/openjdk/jdk8/blob/master/jdk/src/share/classes/java/io/OutputStream.java#L74) it is.

spec/java/lang/System.main.lsl
dog-m: *Technically* we can write a loop and array type checks there, but it is more efficient to copy symbolic memory around using intrinsics of the USVM (none of which I am aware of).

spec/java/lang/System.main.lsl
dog-m: Static string messages do not have such issues.

spec/java/util/ArrayList.SubList.ListIterator.lsl
dog-m: Sorry, but I think it is unnecessary due to this being the only place where `_throwNPE` can be used.

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
dog-m:
```
jshell> var a = 5; a == true
a ==> 5
|  Error:
|  incomparable types: int and boolean
|   a == true
|   ^-------^
```

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
dog-m:
```
jshell> int x = true
|  Error:
|  incompatible types: boolean cannot be converted to int
|  int x = true;
|          ^--^
```

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
dog-m:
```
jshell> var a = 5; var b = true; a = b
a ==> 5
b ==> true
|  Error:
|  incompatible types: boolean cannot be converted to int
|   a = b
|       ^
```

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
dog-m:
```
jshell> var a = 5; var b = true; a = b
a ==> 5
b ==> true
|  Error:
|  incompatible types: boolean cannot be converted to int
|   a = b
|       ^
```

spec/java/util/concurrent/atomic/AtomicBoolean.main.lsl
dog-m: If it is *possible* then we *should* __model__ the expected behavior. But we *can't* delay the assignment.

The only way I can think of is if we use `SYMBOLIC(boolean)`. But in that case we can end up having an infinite assignment loops in user programs when considered multi-threading environments (which is still unsupported).

spec/java/util/stream/DoubleStream.lsl
dog-m: I don't see the necessity of it.
I suggest changing it later when we will have more than 2 places for that.
It also not that difficult to just write as individual constants.

spec/java/util/stream/DoubleStream.lsl
dog-m: I'm sorry, but I can't quite parse what you mean by that. Is this a request for changes or just a side-note?

spec/runtime/utils/SymbolicInputStream.lsl
dog-m: It is a utility class. I can move it into `libsl.utils.*` package for better clarity if needed.

spec/runtime/utils/VoidInputStream.lsl
dog-m: *"Null"* usually means *"nothing"*. *"Void"*, on the other hand, is *"empty"*, if I'm not mistaken.
So, this is an *"empty input stream"*.

spec/runtime/utils/VoidOutputStream.main.lsl
dog-m: [Yes](https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/io/OutputStream.java#L122).

spec/java/lang/ClassLoader.lsl
dog-m: Fixed

spec/java/lang/System.StdOut.lsl
dog-m: Fixed

spec/java/lang/System.main.lsl
dog-m: Fixed

spec/java/lang/Double.main.lsl
alexeev509: @dog-m Why we can't write such checking: "v == -0.0" ? Or LibSL can't work with such construction ?

spec/java/lang/SecurityManager.main.lsl
alexeev509: @dog-m Maybe add comment about this ?

spec/java/util/stream/DoubleStream.lsl
alexeev509: I suggest to back this spec to package "java.util.DoubleStreamLSLSpliterator"
because original class url is in package: /**java/util**/Spliterators.java

spec/runtime/utils/SymbolicInputStream.lsl
alexeev509: @dog-m Yes, this name will be better )

spec/libsl/utils/VoidInputStream.lsl
```diff
@@ -16,7 +16,7 @@ import java/io/InputStream;
 @GenerateMe
 @extends("java.io.InputStream")
 @public @final type VoidInputStream
-    is java.io.InputStream$Void
+    is `libsl.utils.VoldInputStream`
```
alexeev509: Void*

dog-m: Fixed

alexeev509 (approved):

PR #23, opened 2023-11-16
Spec for `java.security.SecureRandom`

alexeev509:
Was created spec for SecureRandom class;
(I'm not sure in realization of the method "getInstanceStrong" )


spec/java/security/Provider.lsl
```diff
@@ -0,0 +1,27 @@
...
+
+
+// local semantic types
+
+@extends("java.util.Properties")
+@abstract type Provider
+    is java.security.Provider
+    for Properties
+{
+    @private @static @final var serialVersionUID: long = -4298000515446427739L;
```
dog-m:
```suggestion
    @private @static val serialVersionUID: long = -4298000515446427739L;
```

spec/java/security/Provider.lsl
```diff
@@ -0,0 +1,27 @@
...
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/Provider.java";
+
+// imports
+
+import java/util/Properties;
+import java/lang/String;
+import java/util/Set;
+
+
+// local semantic types
```
dog-m: Please make distinct sections using comments:
```scala
// primary semantic types
...
<declaration for the original type here>
...
<globally-accessible constants here>
...
// global aliases and type overrides
...
<approximation-specific "extension" type here>
```

spec/java/security/SecureRandom.lsl
```diff
@@ -0,0 +1,22 @@
...
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/SecureRandom.java";
+
+// imports
+
+import java/util/Random;
+
+
+// local semantic types
```
dog-m: Please make distinct sections using comments and use separate "mirror" type declaration, like [this one](https://github.com/vpa-research/jsl-spec/blob/patches-mikhail/spec/java/lang/Float.lsl).

spec/java/security/SecureRandom.lsl
```diff
@@ -0,0 +1,22 @@
...
+
+@extends("java.util.Random")
+@public type SecureRandom
+    is java.security.SecureRandom
+    for Random
+{
+    @private @static @final var serialVersionUID: long = 4940670005562187L;
+    @private @static var defaultProvidersMap: map<String, Object> = action MAP_NEW();
```
dog-m: If this map (which is actually a *set*) reference is not being updated anywhere, I suggest this:
```suggestion
    @private @static val defaultProvidersMap: map<String, Object> = action MAP_NEW();
```

spec/java/security/SecureRandomSpi.lsl
```diff
@@ -0,0 +1,22 @@
...
+import java/io/Serializable;
+import java/lang/Object;
+
+
+// local semantic types
+
+@implements("java.io.Serializable")
+@abstract type SecureRandomSpi
+    is java.security.SecureRandomSpi
+    for Object, Serializable
```
dog-m: `Serializable` is also an `Object` so no need referencing `Object` here, I guess.

spec/java/security/Security.lsl
```diff
@@ -0,0 +1,23 @@
...
+
+// imports
+
+import java/lang/Object;
+import java/security/Provider;
+
+
+// local semantic types
+
+@public @final type Security
```
dog-m: `@public` is not needed here.

spec/java/security/Service.lsl
```diff
@@ -0,0 +1,23 @@
...
+
+// imports
+
+import java/lang/Object;
+import java/lang/String;
+
+
+// local semantic types
+
+@public @static type Provider_Service
```
dog-m: Why `Provider.Service` is declared here, but not in the same file where the `Provider` is declared?

Also `@public` is not needed here.

spec/java/security/SecureRandomParameters.lsl
```diff
@@ -0,0 +1,19 @@
...
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/SecureRandomParameters.java";
+
+// imports
+
+import java/lang/Object;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/security/SecureRandomParameters.lsl
```diff
@@ -0,0 +1,19 @@
+libsl "1.1.0";
+
+library std
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/SecureRandomParameters.java";
```
dog-m: Be ware, this file exists [only since Java 9](https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/SecureRandomParameters.java#L36). This can cause problems during compilation of translated sources.

spec/java/security/Security.lsl
```diff
@@ -0,0 +1,23 @@
...
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/Security.java";
+
+// imports
+
+import java/lang/Object;
+import java/security/Provider;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/util/Dictionary.lsl
```diff
@@ -0,0 +1,19 @@
...
+
+import java/lang/Object;
+
+
+// local semantic types
+
+@abstract type Dictionary
+    is java.util.Dictionary
+    for Object
+{
```
dog-m, line 20: I suggest adding the following set of method **declarations** here:
- `size`
- `isEmpty`
- `get`
- `put`
- `remove`
- `keys`
- `elements`
*(which is all of them?)*

spec/java/security/Service.lsl
```diff
@@ -0,0 +1,23 @@
...
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/Provider.java";
+
+// imports
+
+import java/lang/Object;
+import java/lang/String;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/security/SecureRandomSpi.lsl
```diff
@@ -0,0 +1,22 @@
...
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/security/SecureRandomSpi.java";
+
+// imports
+
+import java/io/Serializable;
+import java/lang/Object;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/util/Dictionary.lsl
```diff
@@ -0,0 +1,19 @@
...
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/util/Dictionary.java";
+
+// imports
+
+import java/lang/Object;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/util/Hashtable.lsl
```diff
@@ -0,0 +1,27 @@
...
+
+@extends("java.util.Dictionary")
+@implements("java.util.Map")
+@implements("java.lang.Cloneable")
+@implements("java.io.Serializable")
+@public type Hashtable
+    is java.util.Hashtable
+    for Dictionary, Map, Cloneable, Serializable
+{
+    @private @static @final var serialVersionUID: long = 1421746759512286392L;
```
dog-m: Since there is no implementation, I suggest removing this field from this type declaration.

spec/java/util/Hashtable.lsl
```diff
@@ -0,0 +1,27 @@
...
+
+// imports
+
+import java/io/Serializable;
+import java/lang/Cloneable;
+import java/util/Dictionary;
+import java/util/Map;
+
+
+// local semantic types
```
dog-m: This is not a `local` type declaration but `// primary semantic types`.

spec/java/util/Properties.lsl
```diff
@@ -0,0 +1,21 @@
...
+    version "11"
+    language "Java"
+    url "https://github.com/openjdk/jdk11/blob/master/src/java.base/share/classes/java/util/Properties.java";
+
+// imports
+
+import java/util/Hashtable;
+
+
+// local semantic types
```
dog-m: I suggest overwriting the contents of this file with [this](https://github.com/vpa-research/jsl-spec/blob/patches-mikhail/spec/java/util/Properties.lsl).

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+// automata
+
+automaton SecureRandomAutomaton
+(
+    var provider: Provider,
+    var algorithm: String,
+    var defaultProvider: boolean
+)
+: SecureRandom
```
dog-m: I suggest, if possible, using LibSL-specific "mirror" type, like it is being done [here](https://github.com/vpa-research/jsl-spec/blob/patches-mikhail/spec/java/lang/Float.main.lsl#L21).

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    proc _getDefaultPRNG (setSeed: boolean, seed: array<byte>): void
+    {
+        val providersList: array<Provider> = action CALL_METHOD(null as Security, "getProviders", []);
+        val providersListLength: int = action ARRAY_SIZE(providersList);
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, providersListLength, +1,
+            findProvider_loop(i, providersList)
```
dog-m:
```suggestion
            _getDefaultPRNG_findProvider(i, providersList)
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @Phantom proc findProvider_loop (i: int, providersList: array<Provider>): void
+    {
+        val curProvider: Provider = providersList[i];
+        val services: Set = action CALL_METHOD(curProvider, "getServices", []);
+        val iter: Iterator = action CALL_METHOD(services, "iterator", []);
+
+        action LOOP_WHILE(
+            action CALL_METHOD(iter, "hasNext", []),
+            findService_loop(iter, curProvider)
```
dog-m:
```suggestion
            _getDefaultPRNG_findProvider_findService(iter, curProvider)
```
or
```suggestion
            _getDefaultPRNG_findService(iter, curProvider)
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @AutoInline @Phantom proc _throwNSPE (): void
+    {
+        action THROW_NEW("java.security.NoSuchProviderException", []);
+    }
+
+
+    proc _getDefaultPRNG (setSeed: boolean, seed: array<byte>): void
+    {
+        val providersList: array<Provider> = action CALL_METHOD(null as Security, "getProviders", []);
```
dog-m: Calling `java.security.Security#getProviders()` will cause USVM to crash because this will eventually call `java.lang.Thread#currentThread()` which is a `native` (unknowable) method.

Also, is this OK to trigger class initialization for `Security` when it makes calls to `File` *(`Unsafe` in `static` [block](https://github.com/openjdk/jdk8/blob/master/jdk/src/share/classes/java/io/File.java#L2181))* and `FileSystem` *([platform](https://github.com/openjdk/jdk8/blob/master/jdk/src/solaris/classes/java/io/DefaultFileSystem.java)-[dependent](https://github.com/openjdk/jdk8/blob/master/jdk/src/windows/classes/java/io/DefaultFileSystem.java))*?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        );
+    }
+
+
+    @Phantom proc findService_loop (iter: Iterator, curProvider: Provider): void
+    {
+        val curService: Provider_Service = action CALL_METHOD(iter, "next", []) as Provider_Service;
+        val curServiceType: String = action CALL_METHOD(curService, "getType", []);
+
+        if (action OBJECT_EQUALS(curServiceType, "SecureRandom"))
```
dog-m: I suggest the other way around to prevent NPE here:
```suggestion
        if (action OBJECT_EQUALS("SecureRandom", curServiceType))
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        }
+    }
+
+
+    @static proc _isDefaultProvider (curProvider: Provider): boolean
+    {
+        val providerName: String = action CALL_METHOD(curProvider, "getName", []);
+
+        if (action MAP_HAS_KEY(this.defaultProvidersMap, providerName))
+            result = true;
```
dog-m: Where is the other option for `result`? Please do not rely on *implicit defaults*.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+
+    proc _nextBytes (bytes: array<byte>): void
+    {
+        // #question: is there a more efficient way?
```
dog-m: Yes: `SYMBOLIC_ARRAY` + `ARRAY_COPY`, I guess.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        action LOOP_FOR(
+            i, 0, size, +1,
+            nextBytes_loop(i, bytes)
+        );
+    }
+
+
+    @static proc _generateSeed (result: array<byte>, numBytes: int): void
+    {
+        // #question: is there a more efficient way?
```
dog-m: `SYMBOLIC_ARRAY` + `ARRAY_COPY`, I guess.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        val size: int = action ARRAY_SIZE(bytes);
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, size, +1,
+            nextBytes_loop(i, bytes)
+        );
+    }
+
+
+    @static proc _generateSeed (result: array<byte>, numBytes: int): void
```
dog-m: What is the key difference between `_generateSeed` and `_nextBytes`?
To me both of them can be combined into one or replaced one by the other.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            i, 0, size, +1,
+            nextBytes_loop(i, bytes)
+        );
+        result = bytes;
+    }
+
+
+    proc _generateRandomIntegerArray (size: int): array<int>
+    {
+        result = action ARRAY_NEW("int", size);
```
dog-m: Why not `SYMBOLIC_ARRAY`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @Phantom proc generateIntArray_loop (i: int, result: array<int>): void
+    {
+        result[i] = action SYMBOLIC("int");
+    }
+
+
+    proc _generateRandomIntegerArrayWithBounds (size: int, randomNumberOrigin: int, randomNumberBound: int): array<int>
+    {
+        result = action ARRAY_NEW("int", size);
```
dog-m: Why not `SYMBOLIC_ARRAY` without individual calls to `SYMBOLIC`, just `ASSYME`s?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        result[i] = action SYMBOLIC("int");
+        action ASSUME(result[i] >= randomNumberOrigin);
+        action ASSUME(result[i] < randomNumberBound);
+    }
+
+
+    proc _generateRandomLongArray (size: int): array<long>
+    {
+        result = action ARRAY_NEW("long", size);
```
dog-m: Why not `SYMBOLIC_ARRAY`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @Phantom proc generateLongArray_loop (i: int, result: array<long>): void
+    {
+        result[i] = action SYMBOLIC("long");
+    }
+
+
+    proc _generateRandomLongArrayWithBounds (size: int, randomNumberOrigin: long, randomNumberBound: long): array<long>
+    {
+        result = action ARRAY_NEW("long", size);
```
dog-m: Why not `SYMBOLIC_ARRAY` without individual calls to `SYMBOLIC`, just `ASSYME`s?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        result[i] = action SYMBOLIC("long");
+        action ASSUME(result[i] >= randomNumberOrigin);
+        action ASSUME(result[i] < randomNumberBound);
+    }
+
+
+    proc _generateRandomDoubleArray (size: int): array<double>
+    {
+        result = action ARRAY_NEW("double", size);
```
dog-m: Why not `SYMBOLIC_ARRAY`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @Phantom proc generateDoubleArray_loop (i: int, result: array<double>): void
+    {
+        result[i] = action SYMBOLIC("double");
+    }
+
+
+    proc _generateRandomDoubleArrayWithBounds (size: int, randomNumberOrigin: double, randomNumberBound: double): array<double>
+    {
+        result = action ARRAY_NEW("double", size);
```
dog-m: Why not `SYMBOLIC_ARRAY` without individual calls to `SYMBOLIC`, just `ASSYME`s?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            generateDoubleArrayWithBounds_loop(i, result, randomNumberOrigin, randomNumberBound)
+        );
+    }
+
+
+    @Phantom proc generateDoubleArrayWithBounds_loop (i: int, result: array<double>, randomNumberOrigin: double, randomNumberBound: double): void
+    {
+        result[i] = action SYMBOLIC("double");
+        action ASSUME(result[i] >= randomNumberOrigin);
+        action ASSUME(result[i] < randomNumberBound);
```
dog-m: Can any element be `NaN`? Should we explicitly state/assume that?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+
+    // special: static initialization
```
dog-m: Please move static class initialization after regular methods, like in [Object](https://github.com/vpa-research/jsl-spec/blob/patches-mikhail/spec/java/lang/Object.main.lsl#L123).

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+
+    @AutoInline @Phantom proc _throwNSPE (): void
+    {
+        action THROW_NEW("java.security.NoSuchProviderException", []);
+    }
+
+
+    proc _getDefaultPRNG (setSeed: boolean, seed: array<byte>): void
```
dog-m: What's the purpose of parameters `setSeed` and `seen` in this subroutine?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // #question: do we need such realization of constructor ? Or it must be empty because this is "private" ?
+    @private constructor *.SecureRandom (@target self: SecureRandom, secureRandomSpi: SecureRandomSpi, provider: Provider, algorithm: String)
+    {
+        action ERROR("Private constructor call");
+    }
+
+
+    constructor *.SecureRandom (@target self: SecureRandom, seed: array<byte>)
+    {
+        _getDefaultPRNG(false, null);
```
dog-m: If `_getDefaultPRNG` subroutine *have* a parameter `seed` then why not use the provided value? Or it is unnecessary in the *first* place?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    }
+
+
+    // static methods
+
+    @throws(["java.security.NoSuchAlgorithmException"])
+    @static fun *.getInstance (_algorithm: String): SecureRandom
+    {
+        val providersList: array<Provider> = action CALL_METHOD(null as Security, "getProviders", []);
```
dog-m: If this is the same implementation as in `_getDefaultPRNG` then why not re-use it by making that subroutine `@static`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+    @throws(["java.security.NoSuchAlgorithmException"])
+    @static fun *.getInstance (_algorithm: String, provider: Provider): SecureRandom
+    {
+        if (provider == null)
+            _throwIAE();
+
+        val providersListLength: int = 1;
+        val providersList: array<Provider> = action ARRAY_NEW("java.security.Provider", providersListLength);
+        providersList[0] = provider;
```
dog-m: Why not check `provider`'s services directly?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+        val providersList: array<Provider> = action CALL_METHOD(null as Security, "getProviders", []);
+        val providersListLength: int = action ARRAY_SIZE(providersList);
+        var resultProvider: Provider = null;
+        var resultAlgorithm: String = null;
+
+        var i: int = 0;
+        action LOOP_FOR(
+            i, 0, providersListLength, +1,
+            findAlgorithmByAllProvidersByName_loop(i, providersList, _algorithm, resultProvider, resultAlgorithm, providerName)
```
dog-m: Looks too complicated to me. Why not just look for the `Provider` and then retrieve the algorithm? Can there be multiple providers with the same name? If so, consider adding a comment on that fact.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+        result = new SecureRandomAutomaton(state = Initialized,
+            provider = resultProvider,
+            algorithm = resultAlgorithm,
+            defaultProvider = isDefaultProvider,
+        );
+    }
+
+
+    @Phantom proc findAlgorithmByAllProvidersByName_loop (i: int, providersList: array<Provider>, algorithm: String, resultProvider: Provider, resultAlgorithm: String, providerName: String): void
```
dog-m:
```suggestion
    @Phantom proc findAlgorithmByAllProvidersByName_loop (
            i: int, providersList: array<Provider>, providerName: String, algorithm: String,
            resultProvider: Provider, resultAlgorithm: String
    ): void
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (propertyLength == 0)
+            _throwNSAE();
+
+        val algorithms: array<String> = action CALL_METHOD(property, "split", [","]);
+        val algorithmsLength: int = action ARRAY_SIZE(algorithms);
+
+        var j: int = 0;
+        action LOOP_FOR(
+            j, 0, algorithmsLength, +1,
+            findAlg_loop(j, algorithms, result)
```
dog-m:
```suggestion
            getInstanceStrong_findAlg(j, algorithms, result)
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    @Phantom proc findAlg_loop (j: int, algorithms: array<String>, result: SecureRandom): void
+    {
+        val curEntry: String = algorithms[j];
+        val algAndProv: array<String> = action CALL_METHOD(curEntry, "split", [":"]);
+        val algAndProvLength: int = action ARRAY_SIZE(algAndProv);
+        val _algorithm: String = algAndProv[0];
+
+        var resultProvider: Provider = null;
+        var resultAlgorithm: String = null;
+        val providersList: array<Provider> = action CALL_METHOD(null as Security, "getProviders", []);
```
dog-m: I suggest moving this call outside of the loop.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        }
+    }
+
+
+    @static fun *.getSeed (numBytes: int): array<byte>
+    {
+        if (numBytes < 0)
+            _throwIAE();
+
+        _generateSeed(result, numBytes);
```
dog-m: `result` is `null` and `_generateSeed` is not being `@AutoInline`d here.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom): DoubleStream
+    {
+        val mass: array<double> = _generateRandomDoubleArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // methods
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom): DoubleStream
+    {
+        val mass: array<double> = _generateRandomDoubleArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomDoubleArray` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, randomNumberOrigin: double, randomNumberBound: double): DoubleStream
+    {
+        val mass: array<double> = _generateRandomDoubleArrayWithBounds(MAX_RANDOM_STREAM_SIZE, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomDoubleArrayWithBounds` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<double> = _generateRandomDoubleArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<double> = _generateRandomDoubleArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom): IntStream
+    {
+        val mass: array<int> = _generateRandomIntegerArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, randomNumberOrigin: int, randomNumberBound: int): IntStream
+    {
+        val mass: array<int> = _generateRandomIntegerArrayWithBounds(MAX_RANDOM_STREAM_SIZE, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<int> = _generateRandomIntegerArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<int> = _generateRandomIntegerArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom): LongStream
+    {
+        val mass: array<long> = _generateRandomLongArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<long> = _generateRandomLongArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, randomNumberOrigin: long, randomNumberBound: long): LongStream
+    {
+        val mass: array<long> = _generateRandomLongArrayWithBounds(MAX_RANDOM_STREAM_SIZE, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<long> = _generateRandomLongArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
```
dog-m: Why not call `LIST_NEW` here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    // within java.util.Random
+    fun *.nextLong (@target self: SecureRandom): long
+    {
+        result = action SYMBOLIC("long");
+    }
+
+
+    @synchronized fun *.setSeed (@target self: SecureRandom, seed: array<byte>): void
+    {
+        action DO_NOTHING();
```
dog-m: Should we check `seed` parameter for being `null`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        action ASSUME(0.0f <= result);
+        action ASSUME(result < 1.0f);
+    }
+
+
+    // within java.util.Random
+    @synchronized fun *.nextGaussian (@target self: SecureRandom): double
+    {
+        result = action SYMBOLIC("double");
+        val isNaN: boolean = action DEBUG_DO("Double.isNaN(result)");
```
dog-m:
```suggestion
        val isNaN: boolean = result != result;
```

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, streamSize: long): DoubleStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 368: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<double> = _generateRandomDoubleArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomDoubleArray` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, streamSize: long, randomNumberOrigin: double, randomNumberBound: double): DoubleStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 388: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, streamSize: long): IntStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 452: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, streamSize: long, randomNumberOrigin: int, randomNumberBound: int): IntStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 472: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, streamSize: long): LongStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 501: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, streamSize: long, randomNumberOrigin: long, randomNumberBound: long): LongStream
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
```
dog-m, line 534: I strongly suggest adding a `// WARNING: ...` comment here about this (our own) extra constraint.

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, streamSize: long): DoubleStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 364: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, streamSize: long, randomNumberOrigin: double, randomNumberBound: double): DoubleStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 382: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, streamSize: long): IntStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 448: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, streamSize: long, randomNumberOrigin: int, randomNumberBound: int): IntStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 466: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, streamSize: long): LongStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 497: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, streamSize: long, randomNumberOrigin: long, randomNumberBound: long): LongStream
+    {
+        var size: int = streamSize as int;
```
dog-m, line 528: What if `size` is negative?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, streamSize: long, randomNumberOrigin: long, randomNumberBound: long): LongStream
```
dog-m, line 526: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, randomNumberOrigin: long, randomNumberBound: long): LongStream
```
dog-m, line 513: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, streamSize: long, randomNumberOrigin: int, randomNumberBound: int): IntStream
```
dog-m, line 464: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, randomNumberOrigin: int, randomNumberBound: int): IntStream
```
dog-m, line 433: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = size,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, streamSize: long, randomNumberOrigin: double, randomNumberBound: double): DoubleStream
```
dog-m, line 380: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
+            length = MAX_RANDOM_STREAM_SIZE,
+            closeHandlers = emptyCloseHandlersList,
+        );
+    }
+
+
+    // within java.util.Random
+    fun *.doubles (@target self: SecureRandom, randomNumberOrigin: double, randomNumberBound: double): DoubleStream
```
dog-m, line 349: What if `randomNumberBound` is **less** than `randomNumberOrigin`?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom): IntStream
+    {
+        val mass: array<int> = _generateRandomIntegerArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomIntegerArray` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<double> = _generateRandomDoubleArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new DoubleStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomDoubleArrayWithBounds` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+
+    // within java.util.Random
+    fun *.ints (@target self: SecureRandom, randomNumberOrigin: int, randomNumberBound: int): IntStream
+    {
+        val mass: array<int> = _generateRandomIntegerArrayWithBounds(MAX_RANDOM_STREAM_SIZE, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomIntegerArrayWithBounds` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<int> = _generateRandomIntegerArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomIntegerArray` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<int> = _generateRandomIntegerArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new IntStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomIntegerArrayWithBounds` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom): LongStream
+    {
+        val mass: array<long> = _generateRandomLongArray(MAX_RANDOM_STREAM_SIZE);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomLongArray` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<long> = _generateRandomLongArray(size);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomLongArray` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+
+
+    // within java.util.Random
+    fun *.longs (@target self: SecureRandom, randomNumberOrigin: long, randomNumberBound: long): LongStream
+    {
+        val mass: array<long> = _generateRandomLongArrayWithBounds(MAX_RANDOM_STREAM_SIZE, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomLongArrayWithBounds` directly here?

spec/java/security/SecureRandom.main.lsl
```diff
@@ -0,0 +1,885 @@
...
+    {
+        var size: int = streamSize as int;
+        if (size > MAX_RANDOM_STREAM_SIZE)
+            size = MAX_RANDOM_STREAM_SIZE;
+
+        val mass: array<long> = _generateRandomLongArrayWithBounds(size, randomNumberOrigin, randomNumberBound);
+        var emptyCloseHandlersList: list<Runnable> = action LIST_NEW();
+
+        result = new LongStreamAutomaton(state = Initialized,
+            storage = mass,
```
dog-m: Why not call `_generateRandomLongArrayWithBounds` directly here?

dog-m (changes requested):

spec/java/security/Provider.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.lsl
alexeev509: Fixed

spec/java/security/SecureRandomSpi.lsl
alexeev509: Fixed

spec/java/security/Security.lsl
alexeev509: Fixed

spec/java/security/Service.lsl
alexeev509: Fixed

spec/java/security/SecureRandomParameters.lsl
alexeev509: Fixed

spec/java/security/SecureRandomParameters.lsl
alexeev509: Fixed (it was deleted);

spec/java/security/Security.lsl
alexeev509: Fixed

spec/java/security/Service.lsl
alexeev509: Fixed

spec/java/security/SecureRandomSpi.lsl
alexeev509: Fixed

spec/java/util/Dictionary.lsl
alexeev509: Fixed

spec/java/util/Hashtable.lsl
alexeev509: @dog-m delete it completely ?

spec/java/util/Hashtable.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m I think it must have such suffix: "loop"; I have fixed in such way: "_getDefaultPRNG_findProvider_loop"

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m @dog-m I think it must have such suffix: "loop"; I have fixed in such way: "_getDefaultPRNG_findService_loop"

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/Provider.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.lsl
alexeev509: Fixed

spec/java/util/Properties.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: As I understood you correctly you want me to add such lines:

```
        if (action MAP_HAS_KEY(this.defaultProvidersMap, providerName))
            result = true;
        else
            result = false;
```

But why I can't rely on default "false" value of the result field ?

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m I suppose it isn't necessary and this param can be deleted from proc _getDefaultPRNG;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
Result was fixed; 

Why it must be "@AutoInline" ?

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m As I understand - no;

```
    /**
     * Reseeds this random object. The given seed supplements, rather than
     * replaces, the existing seed. Thus, repeated calls are guaranteed
     * never to reduce randomness.
     *
     * @param seed the seed.
     *
     * @see #getSeed
     */
    synchronized public void setSeed(byte[] seed) {
        secureRandomSpi.engineSetSeed(seed);
    }
```

For example in secureRandomSpi can be such code:
```
if (seed == null)
    generateOwnSpecialSeed();
```

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/util/Dictionary.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: Loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: Loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: Were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

Fixed;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
It has some differences; It doesn't throw exception if provider equals null;

It throws another exception if algorithm equals null;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
These lines were deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
This loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
This loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
This loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/security/SecureRandom.main.lsl
alexeev509: @dog-m 
This loop was deleted, because were used:

```
action SYMBOLIC("java.security.Provider");
action SYMBOLIC("java.lang.String");
```

temporary decision;

spec/java/util/Random.main.lsl
```diff
@@ -63,6 +63,74 @@ automaton RandomAutomaton
...
+        action LOOP_FOR(
+            i, 0, size, +1,
+            checkDoubleBounds_loop(i, result, randomNumberOrigin, randomNumberBound)
+        );
+    }
+
+
+    @Phantom proc checkDoubleBounds_loop (i: int, result: array<double>, randomNumberOrigin: double, randomNumberBound: double): void
+    {
+        action ASSUME(result[i] != DOUBLE_NAN);
```
dog-m: Incorrect NaN check for Java translation target.
Please use this instead:
```dart
action ASSUME(result[i] == result[i]);
```

Also, consider making a local variable and then bind constraints on that:
```dart
val item: double = result[i];
action ASSUME(item == item);
...
```

dog-m (changes requested):

spec/java/security/SecureRandom.main.lsl
dog-m: Why not directly like this:
```dart
result = action MAP_HAS_KEY(...);
```
?

spec/java/security/SecureRandom.main.lsl
dog-m: As I can see it from existing implementations like:
- `sun.security.provider.SecureRandom#engineSetSeed` or
- `sun.security.pkcs11.P11SecureRandom#engineSetSeed`

it *will* cause NullPointerException to be thrown.

dog-m (approved):

PR #24, opened 2023-11-21
Fixes for `java.lang.StringBuffer` and `java.lang.StringBuilder`

kechinvv:


kechinvv:
I added to StrinBuilder same fixes as in StringBuffer. But the methods that failed the tests with NPE were not fixed because I'm not sure there is a problem with them (But in StringBuffer I tried to fix them).

dog-m (approved):

PR #29, opened 2024-01-20
Re-implementation of `java.lang.StringBuffer/StringBuilder`

dog-m:
Key changes:
- Reworked implementation of `java.lang.StringBuffer`.
- Reworked implementation of `java.lang.StringBuilder`.
- Added declaration for action `ARRAY_FILL_RANGE`.
- Added *partial* implementation (patches) for `java.lang.Character` and `java.lang.String`.

### DIFF
--- a/spec/java/lang/StringBuffer.lsl
+++ b/spec/java/lang/StringBuffer.lsl
@@ -18,14 +18,11 @@ import java/lang/Comparable;
 @implements("java.io.Serializable")
 @implements("java.lang.Comparable<StringBuffer>")
 @implements("java.lang.CharSequence")
-
 @final type StringBuffer
     is java.lang.StringBuffer
     for Serializable, Appendable, Comparable, CharSequence
 {
     @private @static val serialVersionUID: long = 3388685877147921107L;
-
-    @private @static val STRINGBUFFER_LENGTH_MAX: int = 50;
 }
 
 
@@ -39,4 +36,6 @@ import java/lang/Comparable;
     for StringBuffer
 {
     @private @static val serialVersionUID: long = 3388685877147921107L;
+
+    @private @static val STRINGBUFFER_LENGTH_MAX: int = 50;
 }
--- a/spec/java/lang/StringBuffer.lsl
+++ b/spec/java/lang/StringBuffer.lsl
@@ -1,4 +1,3 @@
-//#! pragma: non-synthesizable
 libsl "1.1.0";
 
 library std
--- a/spec/java/lang/StringBuffer.lsl
+++ b/spec/java/lang/StringBuffer.lsl
@@ -25,6 +25,8 @@ import java/lang/Comparable;
     for Serializable, Appendable, Comparable, CharSequence
 {
     @private @static val serialVersionUID: long = 3388685877147921107L;
+
+    @private @static val STRINGBUFFER_LENGTH_MAX: int = 50;
 }
 
 
--- a/spec/java/lang/StringBuffer.main.lsl
+++ b/spec/java/lang/StringBuffer.main.lsl
@@ -8,8 +8,9 @@ library std
 // imports
 
 import java/lang/Character;
+import java/util/stream/IntStream;
+
 import java/lang/StringBuffer;
-import java/lang/Runnable;
 
 
 // automata
--- a/spec/java/lang/StringBuffer.main.lsl
+++ b/spec/java/lang/StringBuffer.main.lsl
@@ -1,3 +1,4 @@
+//#! pragma: target=java
 libsl "1.1.0";
 
 library std
--- a/spec/java/lang/StringBuffer.main.lsl
+++ b/spec/java/lang/StringBuffer.main.lsl
@@ -1,4 +1,3 @@
-//#! pragma: non-synthesizable
 libsl "1.1.0";
 
 library std
@@ -376,7 +375,8 @@ automaton StringBufferAutomaton
     {
         _preconditionCheck();
 
-        action TODO();
+        val seq: String = action OBJECT_TO_STRING(str);
+        _appendCharSequence(seq, offset, len);
 
         result = self;
     }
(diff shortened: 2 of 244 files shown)


spec/java/lang/StringBuffer.lsl
```diff
@@ -25,6 +24,8 @@ import java/lang/Comparable;
     for Serializable, Appendable, Comparable, CharSequence
 {
     @private @static val serialVersionUID: long = 3388685877147921107L;
+
+    @private @static val STRINGBUFFER_LENGTH_MAX: int = 50;
```
kechinvv: Why max is 50? Is not it too small?

spec/java/lang/StringBuffer.main.lsl
```diff
@@ -1,38 +1,40 @@
...
 
 
 // automata
 
 automaton StringBufferAutomaton
 (
+    var value: array<char> = null, // WARNING: do not rename or change the type!
+    var count: int = 0             // WARNING: do not rename or change the type!
 )
 : StringBuffer
```
kechinvv: it may be my old mistake, but automaton usually return LSL type like LSLStringBuffer

spec/java/lang/StringBuffer.main.lsl
```diff
@@ -91,987 +94,936 @@ automaton StringBufferAutomaton
...
         if (another == self)
         {
             result = 0;
         }
         else
         {
-            val anotherString: String = StringBufferAutomaton(another).storage;
-            result = action CALL_METHOD(this.storage, "compareTo", [anotherString]);
+            val thisString: String = _asString();
+            val anotherString: String = StringBufferAutomaton(another)._asString();
```
kechinvv, line 599: We can call functions via dot?

kechinvv: Also I not saw this construction before, it is valid? We have not this constructor for automaton or it is cast?

spec/java/lang/StringBuffer.main.lsl
```diff
@@ -91,987 +94,936 @@ automaton StringBufferAutomaton
...
-    proc _checkRangeSIOOBE (start: int, end: int, len: int): void
+    @AutoInline @Phantom proc _preconditionCheck (): void
     {
-        if (start < 0 || start > end || end > len)
-        {
-            //val message: String = "start " + action OBJECT_TO_STRING(start) + ", end " + action OBJECT_TO_STRING(end) + ", length " + action OBJECT_TO_STRING(len);
-            action THROW_NEW("java.lang.StringIndexOutOfBoundsException", []);
-        }
-    }
+        // #question: do we need ALL of that?
```
kechinvv, line 107: It is really a duplicate code, but we shouldn't throw concretely StringIndex.. exception?

spec/java/lang/StringBuffer.main.lsl
```diff
@@ -91,987 +94,936 @@ automaton StringBufferAutomaton
...
+    // WARNING: seq should NOT be NULL!
+    proc _appendCharSequence (seq: CharSequence, seqStart: int, seqEnd: int): void
     {
-        result = (codePoint >= MIN_CODE_POINT && codePoint <= MAX_CODE_POINT);
-    }
+        if ((seqStart < 0) || (seqStart > seqEnd) || (seqEnd > action CALL_METHOD(seq, "length", [])))
+            action THROW_NEW("java.lang.IndexOutOfBoundsException", []);
 
+        // trim long strings
+        val avaiable: int = action ARRAY_SIZE(this.value) - this.count;
```
kechinvv, line 123: We have STRINGBUFFER_LENGTH_MAX, should we call ARRAY_SIZE?

spec/java/lang/StringBuffer.lsl
kechinvv: May be we can realloc arrays... But I don't know about usvm's requirements - if it's ok then ok

spec/java/lang/StringBuffer.main.lsl
```diff
@@ -91,987 +94,936 @@ automaton StringBufferAutomaton
...
     }
 
 
     @synchronized fun *.trimToSize (@target self: StringBuffer): void
     {
-        // storage is dynamic, so nothing more
-        action DO_NOTHING();
+        _preconditionCheck();
+
+        // #question: what do we need to do here?
```
kechinvv, line 1023: Before was "action DO_NOTHING", i think it is ok

spec/java/lang/StringBuffer.lsl
dog-m: This is similar to the requested implementation:
* [UtStringBuffer](https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-framework/src/main/java/org/utbot/engine/overrides/strings/UtStringBuffer.java#L27)
* [Resolver](https://github.com/UnitTestBot/UTBotJava/blob/main/utbot-framework/src/main/kotlin/org/utbot/engine/Resolver.kt#L117)

I suspect the change itself arises from some internal issues of memory management or constraint solving inside of the USVM.

spec/java/lang/StringBuffer.main.lsl
dog-m: As of now it is a supported expression. It is widely used in `java.util.ArrayList` and its companion classes.
Don't see any issue here.

spec/java/lang/StringBuffer.main.lsl
dog-m: `java.lang.StringIndexOutOfBoundsException` is being thrown elsewhere.

How is this relate to the comment itself?

spec/java/lang/StringBuffer.main.lsl
dog-m: It should be easier for the USVM to connect the actions and make decisions during error discovery / "execution".

I suspect the "error" looks something like this for array access operations:
```java
array[index] = ...

// turns into

if (index < 0 || array.length <= index)
  error("Invalid array index: " + index);
else
  array[index] = ...
```

As you can see, the `length` field is being compared to the provided `index`.

spec/java/lang/StringBuffer.main.lsl
kechinvv: At first it seemed to me that this was a question for the old code. If we talk about "assume" I think they will not be superfluous

spec/java/lang/StringBuffer.main.lsl
dog-m: Fixed.

kechinvv (approved):

